PR #61, opened 2019-06-25
Method: remove injection, pass metadata to constructor instead

axelchalon:
Waiting on https://github.com/polkadot-js/api/pull/1047 (updating the dependency once the api PR is merged)

### DIFF
--- a/packages/extension-ui/src/Popup/Signing/Details.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Details.tsx
@@ -5,7 +5,7 @@
 import React from 'react';
 import styled from 'styled-components';
 import findChain from '@polkadot/extension/chains';
-import { Metadata, Method } from '@polkadot/types';
+import { GenericCall, Metadata } from '@polkadot/types';
 
 type MethodJson = {
   args: { [index: string]: any }
@@ -30,7 +30,7 @@ function renderMethod (data: string, meta?: Metadata | null) {
     );
   }
 
-  const method = new Method(data, { meta });
+  const method = new GenericCall(data, { metadata: meta });
   const json = method.toJSON() as unknown as MethodJson;
 
   return (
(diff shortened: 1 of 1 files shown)


packages/extension-ui/src/Popup/Signing/Details.tsx
```diff
@@ -31,9 +30,7 @@ function renderMethod (data: string, meta?: Metadata | null) {
     );
   }
 
-  Method.injectMethods(fromMetadata(meta));
-
-  const method = new Method(data);
+  const method = new Method(data, { meta });
```
jacogr: Is this correct? meta is not the method metadata, it is everything.

axelchalon: yes, we made it so that the constructor can either accept the function metadata or the whole metadata (from which it extracts the relevant function metadata), so that it's more practical for end-users (developers)

https://github.com/polkadot-js/api/pull/1047#discussion_r297348952

jacogr: Cool, that is nice. Caters for this specific use-case.

packages/extension-ui/src/Popup/Signing/Details.tsx
```diff
@@ -30,7 +30,7 @@ function renderMethod (data: string, meta?: Metadata | null) {
     );
   }
 
-  const method = new Method(data, { meta });
+  const method = new GenericCall(data, { metadata: meta });
```
jacogr: So not sure about this one at all - so the idea with the API is that if this happens, that it should be the full metadata. And then... in steps the extension... it doesn't have any API instance (only types) and maybe it has updated metadata available.

axelchalon: Not sure I understand; `meta` is the full metdata. Currently the full metadata is hard-coded in the extension ui, and passed to the Method/GenericCall constructor. Do we want to change this?

jacogr:
Closing. The API interface has been changed around this by passing the registry around, and using the registry for lookups.

PR #72, opened 2019-07-11
Swap to eslint

jacogr:


packages/extension-ui/src/components/Button.tsx
```diff
@@ -44,54 +44,26 @@ function Button ({ children, className, isDisabled, label, onClick, to }: Props)
 
 export default styled(Button)`
   box-sizing: border-box;
-  display: ${({ isSmall }) =>
-    isSmall
-      ? 'inline-block'
-      : 'block'
-  };
+  display: ${({ isSmall }): string => isSmall ? 'inline-block' : 'block'};
```
jacogr: This is ok, but adjusted due to the next comment...

packages/extension-ui/src/components/Button.tsx
```diff
@@ -44,54 +44,26 @@ function Button ({ children, className, isDisabled, label, onClick, to }: Props)
...
-      isDanger
-        ? defaults.btnBgDanger
-        : defaults.btnBg
-    };
-    border: ${defaults.btnBorder}${({ isDanger }) =>
-      isDanger
-        ? defaults.btnColorDanger
-        : defaults.btnColor
-    };
+    background: ${({ isDanger }): string => isDanger ? defaults.btnBgDanger : defaults.btnBg};
```
jacogr: Ok, this really sucks.

The issue is that eslint complains about the indent, so actually would like this (when multi-line) -

```js
background: ${({ isDanger }) =>
  isDanger
    ? defaults.btnBgDanger
    : defaults.btnBg
```

This is due to the fact that this is nested inside `button { ... }`

PR #86, opened 2019-07-15
Hack to hide scrollbars on popups

### LINKED ISSUE #80: Firefox - Scrollbar on the popup
Any popup on Firefox (welcome message, signing) opens a window with scrollbars.

Removing the height and width to the body css fixes it, not sure that's the best idea (and haven't tested on Chrome yet)

image

Tbaut:
1. closes #80 
Not particularly proud of that one.. yet I've spent some time trying to figure out what happens.. without luck. It must be a Firefox bug.. A single pixel of the height lets a vertical scroll-bar appear, which then lets an horizontal scroll-bar appear (because in FF the scroll-bar is part of the window). So 580 -> 581 solves the whole problem. Had a look at what Metamask does and their extension is 357x600 and they open a 360x620 popup :woman_cartwheeling:.

1. Now the `overflow-x: hidden` solves another scrollbar problem when having many accounts, before
image

after
image

Note that this is totally transparent for chrome..

### DIFF
--- a/packages/extension/src/background/handlers/State.ts
+++ b/packages/extension/src/background/handlers/State.ts
@@ -91,7 +91,7 @@ export default class State {
     extension.windows.create({
       // This is not allowed on FF, only on Chrome - disable completely
       // focused: true,
-      height: 580,
+      height: 581,
       left: 150,
       top: 150,
       type: 'popup',
--- a/packages/extension/public/popup.html
+++ b/packages/extension/public/popup.html
@@ -4,7 +4,7 @@
     <meta charset="utf-8">
     <title>polkadot{.js}</title>
   </head>
-  <body style="background:#f5f6f7; box-sizing:border-box; height:580px; margin:0; padding:0 1rem; width:480px;">
+  <body style="background:#f5f6f7; box-sizing:border-box; height:580px; margin:0; overflow-x: hidden; padding:0 1rem; width:480px;">
     <div id="root"></div>
     <script src='./popup.js'></script>
   </body>


packages/extension/src/background/handlers/State.ts
```diff
@@ -91,7 +91,7 @@ export default class State {
     extension.windows.create({
       // This is not allowed on FF, only on Chrome - disable completely
       // focused: true,
-      height: 580,
+      height: 581,
```
jacogr: yes, well...

Tbaut: The window size doesn't need to remain this ugly, but it seems like it needs to remain > body/html size.
I tried many things like setting the style in the `html` with the body `100%`, with nothing in the body but that vertical scrollbar remains.

jacogr: I understand - the basic issue is that the borders are included in the width/height when making a popup. Just for interest, did you try `box-sizing`?

Happy as it is, since it, well, is what it is :)

Tbaut: I tried just to remove box-sizing which broke the sizing altogether iirc, but didn't play further with it. Will do a last chance play with box-sizing tomorrow.

jacogr: No, it is fine. Was just interested. (i.e. basically had issues like this ages ago, the work-around was similar to what was actually done here - so was just wondering if there are maybe better options. Would not spend too much additional effort)

Issue #58, opened 2019-06-14
Implement Provider for extension injection


jacogr:

axelchalon:
What is the benefit of injecting a PostMessage provider, that sends the requests to the extension, which forwards them to the node (using WsProvider); over injecting the WsProvider directly?

jacogr:
You don't want to run anything `@polkadot/api` related in the UI or injection threads. Actually, if we had to do that atm, you would end up with the dreaded "multiple packages detected" message.

Additionally, you want to be able to have max flexibility on the provider, if you need to -

- provide access for only some time
- kill it immediately at any point at the user's discression
- provide control over methods that are available (i.e. allow queries, no submission)
- share the provider (not applicable for extension, maybe for metadata, but certainly for others)

These are all things that can be done (certainly not suggesting it for a minimal-initial version)

jacogr:
Done.

Issue #85, opened 2019-07-15
Manage access as previously granted


jacogr:
Goal:

The user should be able to check access on a per site basis and either approve/reject these.

Actions:

- storage permissions to storage
- page where the user to reject or approve based on sites that requested access

Similar to #10

jacogr:
#289 May have a simpler (at least first phase, whack-all) solution

tomusdrw:
Would be good to have at least an option to see "recently rejected" websites (even if not persisted) with an option to revert, in case user accidentally rejects the prompt.

Issue #15, opened 2019-05-21
Add settings


jacogr:
Goal:

Make it easy to tweak.

Actions:

- Hamburger
- tweak of icon theme (currently default to polkadot)

For the future (once we actually have something):

- tweak of type, i.e. inject accounts, inject provider + accounts
- network selection

... more?

jacogr:
Done with the redesign.

Issue #10, opened 2019-05-21
Select accounts to be made visible


jacogr:
Goal:

Allow the user better control over the accounts injected.

Background:

Currently all available accounts are injected into the page. Permissions per site are to be set via https://github.com/polkadot-js/extension/issues/9 - but we would like to extend this with account control.

Actions:

Allow the user to select the visible accounts on a global level and allow adjustments thereof on a per-site basis. So by default the defaults will be injected, however per site the accounts can be limited or overridden.

PR #88, opened 2019-07-16
Provider injection

axelchalon:
Close https://github.com/polkadot-js/extension/issues/58

Preparing the groundwork. Next up in this PR: integrating this in `extension`


packages/extension-inject/src/types.ts
```diff
@@ -35,6 +45,7 @@ export interface InjectedExtensionInfo {
 export interface Injected {
   accounts: InjectedAccounts;
   signer: InjectedSigner;
+  provider?: InjectedProvider;
```
jacogr: Just align the docs with this (README right up-front, since it is optional)

axelchalon:
I imagine we also want to add a screen inside the extension where the user can choose the RPC endpoint, if it's different from `ws://127.0.0.1:9944` ? Or is this non-urgent ?

jacogr:
Indeed, https://github.com/polkadot-js/extension/issues/15

We may want to take a peek at that first, at the very least it allows for 2 other quick hits - #10 & #85 

(It is sensical to get that UI work done next, if nobody else grabs it in the meantime I will get around to it once apps is up and humming with 2.x again)

axelchalon:
- Typed sendMessage. TypeScript now checks that the message type is valid, that the payload corresponds to the message type's payload, and that the return type is the correct return type. e.g.
  ```ts
  const foo: boolean = await sendMessage('seed.create', { length: 12, type: 'ed25519' } }); // error, seed.create doesn't return a boolean
  const foo = await sendMessage('seed.create', true }); // error, seed.create expects { length, type }
  ```
- Implement PostMessageProvider: PostMessageProvider uses the regular `sendMessage` function with `method` and `params` as `request`, and gets the response the usual way. For subscription notification handling, we use a separate channel, and PostMessageProvider matches subscription id to handlers.
  Decided not to have PostMessageProvider encode in JSON and pass the JSON to the extension for multiple reasons, among them 1) would need yet another provider (extension would have to use WsRawProvider that forwards json strings to the provider) 2) if we want to implement ACL in the future, the extension would need to parse the incoming JSON of the request, 3) if we want to reuse the current loader architecture for requests/responses it would mean we would have two ids being passed around, the usual sendMessage id & the jsonrpc id, which starts making things confusing

Still have to work out the handling of WsProvider connection/disconnection/reconnection and how it should affect the PostMessageProvider (right now it assumes the extension provider is always on) ; and have to actually test this PR against apps (next step in this PR), but overall I think the gist of it is here. so open for reviewing/comments!


Haven't tested if it works yet, but the gist of it is here. Will test on apps shortly.

packages/extension/src/page/PostMessageProvider.ts
```diff
@@ -0,0 +1,126 @@
+// Copyright 2019 @polkadot/extension authors & contributors
```
jacogr: I have just been scrolling through, so no in-depth comments yet - however, it would be good if we can actually move the provider (as to be injected) to `extension-provider` so it is actually easily re-usable by others as well.

(I can see that this file, as-is, can be shared - just the endpoint can be custom for a "quick start")

jacogr: And yes... need to do _exactly_ the same with the injected signer. (In thinking about these 2, it may end up being in `extension-inject` then as sub-folders.) So extensions do the following -

- include the package
- use the top-level interface to inject into the page
- create an instance of the signer for signing (or do their own)
- create an instance of the provider (or ignore, or do their won)

One self-contained bundle.

axelchalon: > it would be good if we can actually move the provider (as to be injected) to extension-provider so it is actually easily re-usable by others as well.

PostMessageProvider only sends messages to the window, it heavily relies on the logic (and message format) of page.js/content.js so that the messages are sent to the extension and received from the extension; as such I don't think it's very reusable by itself. Same for the injected signer, which basically only does `sendRequest('extrinsic.sign', payload);`. What do you think?

axelchalon:
Prior to the previous commit, the extension was using a single WsProvider, shared by all pages. Since we need active subscriptions made by the page to be closed when the page closes, and since there is no maintainable way of deriving the unsubscription method name from the subscription method name, I resorted to using one WsProvider per page. When the page closes, the associated underlying WsProvider disconnects.

Right now the WsProvider connects when the page makes an RPC request. Connecting or disconnecting the WsProvider from the Page (from the PostMessageProvider) isn't supported as it seems to me like a lot of overhead for little benefit.

packages/extension/src/page/PostMessageProvider.ts
jacogr: I would suggest looking at how other extensions are using this - as it is, it is being used as-is. You do have a good suggestion buried in there though to make the message names slightly less generic for these case, i.e so it is explicit where it comes from and what it is for.

jacogr:
Can we start pulling parts in? i.e. the typing changes - that we can pull in via another PR. (Actually unrelated and will clean this one . up some more). Would like to start getting this ready so it doesn't hang about.

axelchalon:
Sure sounds good, will do this now. I'll go about figuring out a strategy for reusability straight after

jacogr:
Perfect. Don't need to solve all the world's problems at once - happy to get improvements in as they come in. (Not a massive fan . of large PRs hanging about)

yarn.lock
```diff
@@ -2090,14 +2100,28 @@
...
+
 "@polkadot/ts@^0.1.68":
   version "0.1.68"
   resolved "https://registry.yarnpkg.com/@polkadot/ts/-/ts-0.1.68.tgz#ef44eeac95d736ad60dbf83c76acfd6d192c5454"
   integrity sha512-vcnX9cDQJ959QNFlHj7QnKeShBr9GIi4vpVrISaIZpmBuF2d3vO0NMu3Fzn/aNvRFJdmfJoJFX55UdfAvAYC5Q==
   dependencies:
     "@types/chrome" "^0.0.86"
 
-"@polkadot/types@^0.91.0-beta.2":
+"@polkadot/types@^0.91.0-beta.2", "@polkadot/types@^0.91.0-beta.6":
```
jacogr: Would like to avoid this type of thing as much as possible - since `@polkadot/api` already depends on `rpc-provider` we really don't need to pull it in explicitly. (As soon as things start mismatching, we may have issues)

So would suggest reverting these

https://github.com/polkadot-js/extension/pull/88/files#diff-ada64edcbc9054dffd93143b631be6ebR10
https://github.com/polkadot-js/extension/pull/88/files#diff-2536f4ac6a738a37164af87d5b55499bR12

axelchalon: Well, didn't know you could import a dependency's dependencies without having them be in your `package.json`. Learned something new today

:+1:

jacogr:
Closing, not needed, but not being worked on here or merge-able.

PR #97, opened 2019-07-18
Add network prefix settings

jacogr:


jacogr:
I did add some FIXME's to the code. This was to get it working with re-work. As settings expand there needs to be a proper extension-specific setting store (so some jumping . through hoops atm which can be simplifies - i.e. add more code, but make one call to set/get, currently that part is a bit messy)

Issue #103, opened 2019-07-21
Implement new signer.signPayload interface


jacogr:
As introduced in https://github.com/polkadot-js/api/pull/1152 (this also enables v2 transaction support)

PR #106, opened 2019-07-24
Update with signPayload for injected signer

jacogr:
Closes https://github.com/polkadot-js/extension/issues/103


jacogr:
@Tbaut When you have a gap, would just like you send a couple of txs as well - everything seems working in my tests, but just want to make 100% sure.

PR #108, opened 2019-07-24
0.5

jacogr:
@Tbaut A final check (it should be exactly the same as the previous version). Probably worth-while building with `yarn build` as well (since that is how it is built for release)


jacogr:
Thank, I just want to do a final test for my own sanity, but will only do so in the morning - and then push to the stores.

PR #114, opened 2019-08-03
Bump API, support Extrinsics v3

jacogr:
- Bumps other deps as well, closes #111, closes #112 & closes #113


jacogr:
Tested as working.

PR #130, opened 2019-08-29
Add typings to messages

axelchalon:
Typed sendMessage. TypeScript now checks that the message type is valid, that the payload corresponds to the message type's payload, and that the return type is the correct return type. e.g.
  ```ts
  const foo: boolean = await sendMessage('seed.create', { length: 12, type: 'ed25519' } }); // error, seed.create doesn't return a boolean
  const foo = await sendMessage('seed.create', true }); // error, seed.create expects { length, type }
  ```

(WIP as I have to figure out why typescript is complaining, but will be fixed soon)


jacogr:
Currently just seems like some styling issues, i.e. eslint.

jacogr:
Merged in master & fixed the eslint, now I see the tsc errors. Weird ones.

packages/extension/src/background/handlers/Extension.ts
```diff
@@ -195,52 +195,52 @@ export default class Extension {
   }
 
   // eslint-disable-next-line @typescript-eslint/no-explicit-any
-  public async handle (id: string, type: keyof ResponseTypes, request: RequestTypes[typeof type], port: chrome.runtime.Port): Promise<ResponseTypes[typeof type]> {
+  public handle (id: string, type: keyof ResponseTypes, request: RequestTypes[typeof type], port: chrome.runtime.Port): Promise<ResponseTypes[typeof type]> {
     switch (type) {
       case 'authorize.approve':
-        return this.authorizeApprove(request as RequestAuthorizeApprove);
+        return Promise.resolve(this.authorizeApprove(request as RequestAuthorizeApprove));
```
jacogr: Is this extra resolve needed? This is just a lot of boilerplate that async takes care of for you.

axelchalon: I thought the same but eslint was complaining

```
/polkadot-extension/packages/extension/src/background/handlers/Extension.ts                                               
  198:3  error  Async method 'handle' has no 'await' expression  @typescript-eslint/require-await  
```

packages/extension-ui/src/Popup/Create.tsx
```diff
@@ -8,9 +8,7 @@ import { ActionContext, Address, Button, Header, Loading, TextArea } from '../co
 import { createAccount, createSeed } from '../messaging';
 import { Back, Name, Password } from '../partials';
 
-type Props = {};
-
-export default function Create (): React.ReactElement<Props> {
+export default function Create (): React.ReactElement<{}> {
```
jacogr: There . is actually some method behind this madness. When you decide to finally add props, the following can happen -

```js
interface Props {
  className?: string;
}

export default function Create ({ className }: Props): React.ReactElement<{}> {
```

Note how the actual return is still `{}`. The above passes without errors, warnings or linting, but is incorrect. Hence being explicit an always defining `Props` types, even when empty. It is some small protection against future rushing where things could very much get out of whack.

packages/extension-ui/src/messaging.ts
```diff
@@ -96,20 +97,20 @@ export async function createAccount (name: string, password: string, suri: strin
   return sendMessage('accounts.create', { name, password, suri, type });
 }
 
-export async function createSeed (length?: number, type?: KeypairType): Promise<{ address: string; seed: string }> {
+export async function createSeed (length?: 12 | 24, type?: KeypairType): Promise<{ address: string; seed: string }> {
```
jacogr: Probably just need to make a type for these magic numbers. (Should really not change much/often, but easir than tracking then all-over the codebase)

```js
type SeedLengths = 12 | 24;
```

packages/extension/src/background/handlers/Extension.ts
```diff
@@ -29,13 +29,13 @@ export default class Extension {
     this.state = state;
   }
 
-  private accountsCreate ({ name, password, suri, type }: MessageAccountCreate): boolean {
+  private accountsCreate ({ name, password, suri, type }: RequestAccountCreate): boolean {
```
jacogr: The naming here is much better than the original.

packages/extension/src/background/types.ts
```diff
@@ -4,89 +4,137 @@
...
+export type RequestSigningRequests = null;
 
-export interface MessageSeedCreateResponse {
-  address: string;
-  seed: string;
-}
+export type RequestSigningSubscribe = null;
 
-export interface MessageSeedCreate {
+export interface RequestSeedCreate {
```
jacogr: The lengths just below would be adjusted for the `SeedLength` type

packages/extension/src/background/types.ts
```diff
@@ -4,89 +4,137 @@
...
+export type ResponseTypes = {
+  [MessageType in keyof RequestSignatures]: RequestSignatures[MessageType][1]
+};
+
+export interface TransportResponseMessage<TMessageType extends MessageTypes> {
+  error?: string;
+  id: string;
+  response?: ResponseTypes[TMessageType];
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  subscription?: any;
```
jacogr: Which type is `subscription` meant to be?

packages/extension/src/page/index.ts
```diff
@@ -30,14 +29,23 @@ let idCounter = 0;
...
     handlers[id] = { resolve, reject, subscriber };
 
-    window.postMessage({ id, message, origin: 'page', request }, '*');
+    const transportRequestMessage: TransportRequestMessage<TMessageType> = {
+      id,
+      message,
+      origin: 'page',
+      request: request || null as RequestTypes[TMessageType]
+    };
+    window.postMessage(transportRequestMessage, '*');
```
jacogr: (nit) extra blank line before, just splitting var defs and actual actions.

packages/extension-ui/src/Popup/Create.tsx
axelchalon: got it, sounds good!

I actually changed this before merging from master, I was getting the `empty interface` linting error; I just saw you changed `Props` it to `type` :+1:

packages/extension/src/background/handlers/Extension.ts
jacogr: Sorry, only saw that now - yes, the 2.x version now does - in these cases I just silence it since it actually has a valid reason

packages/extension/src/background/types.ts
axelchalon: Good question, it actually doesn't have to be `any` here.

In case of `accounts.subscribe`, `authorize.subscribe`, `signing.subscribe` : `subscription` (subscription message actually) will be respectively of type `KeyringJson[]`, `AuthorizeRequest[]`, `SigningRequest[]`.

Removed this `any` and the one in the `sendMessage` subscription callback (so we have typecheck for the callback now as well)

jacogr: Brilliant, thanks.

axelchalon:
Cool, should I merge then? not sure what the policy is on this

jacogr:
Yes, you can merge at will - unless there are some danglings that you would like to care of. (I generally leave it unmerged, just in case there are some last-minute stuff :))

jacogr:
If you don't get to it, I will probably merge tonight anyway, despite you maybe having some "dangling bits" - would like to get the mobile signer Qr thing ready. (And since I touched some responses and added some messages, there will be some work)

axelchalon:
Sounds good, just need 30 more minutes to wrap up typing the subscription callbacks

jacogr:
No issues, it won't be before 20h30 when things quieten down for me. So lots of time. Just mark WIP if you don't to everything - then I'll probably just merge the branch in mine as-is and mop-up the additionals afterwards.

axelchalon:
So, I typed the subscriptions callback but an incongruency came up: if I'm not mistaken the return value of 'accounts.list'/'accounts.subscribe' is different if handled by the tabs handler (`{address, name}`) or by the extension handler (`{address, meta: {name}} aka KeyringJson`). To make this consistent I made both return `InjectedAccount` i.e. `{address, name}`.  I don't believe this is a breaking change as only the responses sent to the extension UI are affected.

jacogr:
I would rather suggest breaking it into 2 different messages then. So the meta is useful with the current version extension - ie. For the Parity Signer changes the genesisHash is actually used. (In display as well). So it really is for internal identification, where the injected is a limited view. 

The genesisHash is actually a crappy example since it was now added to the injected as well. (However the isExternal on meta is used by the extension part, but it is not exposed to the page)

axelchalon:
pub/pri name splitting is great! :+1:

Issue #122, opened 2019-08-16
Export account


JonasPf:
Hi,

Is there a way to export an account? I'm fine with pulling it out of a db or using an API if it is not possible via the UI.

jacogr:
Not yet, will keep this issue open to add that functionality. You can get to that info using https://chrome.google.com/webstore/detail/storage-area-explorer/ocfjjjjhkpapocigimmppepjgfdecjkb which can inspect the `chrome.storage.local` area.

- navigate to chrome extensions
- view extension background page
- with the extension installed, to to the storage
- the `value` field for the accounts in the encrypted JOSN (i.e. can be imported into e.g. the polkadot-js/apps UI)

JonasPf:
Any tips for Firefox? I got as far as this:

1. Opened `about:debugging`
2. Check `Enable add-on debugging`
3. Click `Debug` next to the extension
4. There seems to be a chrome thing in "Indexed DB"

jacogr:
Closed in #166 (Will be in the 0.13 release)

PR #166, opened 2019-10-01
Export account

Anze1m:
This PR implements functionalities mentioned in https://github.com/polkadot-js/extension/issues/122


jacogr:
Thank you very much, this is really appreciated. I will take a deep-dive into playing and reading through a bit later today.

jacogr:
My appologies, running behind - really want to get to this one ASAP, just haven't caught a breather. Don't want to keep you handling.

jacogr:
Couple of comments on functionality -

- when I type in the wrong password, we should highlight the box - currently nothing happens, with no indication (just a highlight `isError` is ok as the first generation)

- Instead of copying the contents, rather download it - less error-prone

- we allow the export . of external accounts (i.e. as added via QR), this should not be allowed, we don't have the keys nor a password for these (the same will happen when the extension has ledger access, so on . `isExternal`, don't try and export)

Future -

With the 3rd button on the little bar, we probably want to rethink - move "Forget" and "Export" to the left, keep "Edit" on the right since it is probably the option most used (simple flex with space-between . should do the trick for the 2 groups)

__mocks__/extensionizer.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
```
jacogr: Brilliant! We are . seriously lacking in tests overall (i.e. we have . none), this is great to make a start.

Just add a copyright header to this file as well.

packages/extension-ui/package.json
```diff
@@ -10,7 +10,13 @@
     "@polkadot/react-qr": "^0.45.0-beta.14",
     "@types/react-router": "^5.1.0",
     "@types/react-router-dom": "^5.1.0",
+    "@types/sinon-chrome": "^2.2.6",
```
jacogr: This is probably also a dev dependency?

packages/extension-ui/src/Popup/Export.test.tsx
```diff
@@ -0,0 +1,75 @@
+import React from 'react';
```
jacogr: Just add a copyright header, just aligning with the rest.

packages/extension-ui/src/Popup/Export.tsx
```diff
@@ -0,0 +1,67 @@
...
+  };
+
+  return (
+    <div>
+      <Header label='export account' />
+      <Back />
+      <Address address={address}>
+        <Tip header='danger' type='error'>You are exporting your account. Keep it safe and don&apos;t share it with anyone.</Tip>
+        {!passwordEntered && <Input
+          isError={pass.length < MIN_LENGTH}
```
jacogr: As suggested in the other comment, also indicate when the export fails, i.e. we cannot decode.

packages/extension-ui/src/Popup/Export.tsx
```diff
@@ -0,0 +1,67 @@
...
+        />}
+        {!passwordEntered && <Button
+          isDisabled={pass.length === 0}
+          isDanger
+          label='I want to export this account'
+          onClick={_onExportButtonClick}
+          className='export-button'
+          data-export-button
+        />}
+        {passwordEntered && <Tip type='info'>Click below to copy exported JSON to clipboard</Tip>}
```
jacogr: Would move this to download the file.

packages/extension-ui/src/messaging.test.ts
```diff
@@ -0,0 +1,20 @@
+import Adapter from 'enzyme-adapter-react-16';
```
jacogr: (nit) copyright header above

packages/extension/src/background/handlers/Extension.test.ts
```diff
@@ -0,0 +1,28 @@
+import Extension from '@polkadot/extension/background/handlers/Extension';
```
jacogr: (nit) Copyright header

packages/extension/src/background/handlers/Extension.test.ts
```diff
@@ -0,0 +1,28 @@
...
+import keyring from '@polkadot/ui-keyring';
+import { cryptoWaitReady } from '@polkadot/util-crypto';
+import State from '@polkadot/extension/background/handlers/State';
+import { ResponseAccountExport } from '@polkadot/extension/background/types';
+import ExtensionStore from '@polkadot/ui-keyring/stores/Extension';
```
jacogr: (nit) Just ensure these are in alphabetical order

packages/extension-ui/src/Popup/index.tsx
```diff
@@ -16,6 +16,7 @@ import Accounts from './Accounts';
 import Authorize from './Authorize';
 import Create from './Create';
 import Forget from './Forget';
+import Export from './Export';
```
jacogr: (nit) Move one up, alphabetical

jacogr:
Brilliant, thanks for the updates. Will have a final play-through tomorrow.

Anze1m:
Thanks for your appreciation, thanks for your feedback 😁 
I think that I've addressed most of the comments. I've corrected imports, added copyright headers, changed displaying and copying JSON to just downloading it. Furthermore, I've improved password input field behavior and restricted export functionality to be available for non-external accounts only.

PR #196, opened 2019-11-18
Implement signRaw

### LINKED ISSUE #115: Implement signRaw Signer interface
It allows for signing of any data

c410-f3r:
Fixes #115 

With the following interface:

```ts
  inner: SignerPayloadJSON | SignerPayloadRaw;
  sign(pair: KeyringPair): { signature: string };
}
```

`extension` package abstracts the signing procedure using the `sign` method and because `extension-ui` needs to be aware if it is dealing with an extrinsic or a bunch of bytes, `inner` returns the inner payload content of its implementations.

### DIFF
--- a/packages/extension-ui/src/Popup/Signing/Request.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Request.tsx
@@ -3,21 +3,23 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { ExtrinsicPayload } from '@polkadot/types/interfaces';
-import { AccountJson, RequestExtrinsicSign } from '@polkadot/extension/background/types';
+import { AccountJson, RequestSign } from '@polkadot/extension/background/types';
+import { SignerPayloadJSON, SignerPayloadRaw } from '@polkadot/types/types';
 
 import React, { useContext, useState, useEffect } from 'react';
 import { TypeRegistry, createType } from '@polkadot/types';
 
 import { ActionBar, ActionContext, Address, Link } from '../../components';
 import { approveSignPassword, approveSignSignature, cancelSignRequest } from '../../messaging';
-import Details from './Details';
+import Bytes from './Bytes';
+import Extrinsic from './Extrinsic';
 import Qr from './Qr';
 import Unlock from './Unlock';
 
 interface Props {
   account: AccountJson;
   isFirst: boolean;
-  request: RequestExtrinsicSign;
+  request: RequestSign;
   signId: string;
   url: string;
 }
@@ -27,15 +29,17 @@ const registry = new TypeRegistry();
 
 export default function Request ({ account: { isExternal }, isFirst, request, signId, url }: Props): React.ReactElement<Props> | null {
   const onAction = useContext(ActionContext);
-  const [payload, setPayload] = useState<ExtrinsicPayload | null>(null);
+  const [hexBytes, setHexBytes] = useState<string | null>(null);
+  const [extrinsic, setExtrinsic] = useState<ExtrinsicPayload | null>(null);
 
   useEffect((): void => {
-    setPayload(createType(registry, 'ExtrinsicPayload', request, { version: request.version }));
+    const inner = request.inner;
+    if ((inner as SignerPayloadRaw).data) {
+      setHexBytes((inner as SignerPayloadRaw).data);
+    } else {
+      setExtrinsic(createType(registry, 'ExtrinsicPayload', inner, { version: (inner as SignerPayloadJSON).version }));
+    }
   }, [request]);
-
-  if (!payload) {
-    return null;
-  }
 
   const _onCancel = (): Promise<void> =>
     cancelSignRequest(signId)
@@ -49,30 +53,46 @@ export default function Request ({ account: { isExternal }, isFirst, request, si
     approveSignSignature(signId, signature)
       .then((): void => onAction())
       .catch((error: Error) => console.error(error));
-
-  return (
-    <Address
-      address={request.address}
-      genesisHash={request.genesisHash}
-    >
-      {isExternal && isFirst
-        ? <Qr
-          payload={payload}
-          request={request}
-          onSignature={_onSignature}
-        />
-        : <Details
-          isDecoded={isFirst}
-          payload={payload}
-          request={request}
-          url={url}
-        />
-      }
-      {isFirst && !isExternal && <Unlock onSign={_onSign} />}
-      <ActionBar>
-        <div />
-        <Link isDanger onClick={_onCancel}>Cancel</Link>
-      </ActionBar>
-    </Address>
-  );
+  if (extrinsic !== null) {
+    const payload = request.inner as SignerPayloadJSON;
+    return (
+      <Address
+        address={payload.address}
+        genesisHash={payload.genesisHash}
+      >
+        {isExternal && isFirst
+          ? <Qr
+            payload={extrinsic}
+            request={payload}
+            onSignature={_onSignature}
+          />
+          : <Extrinsic
+            isDecoded={isFirst}
+            payload={extrinsic}
+            request={payload}
+            url={url}
+          />
+        }
+        {isFirst && !isExternal && <Unlock onSign={_onSign} />}
+        <ActionBar>
+          <div />
+          <Link isDanger onClick={_onCancel}>Cancel</Link>
+        </ActionBar>
+      </Address>
+    );
+  } else if (hexBytes !== null) {
+    const payload = request.inner as SignerPayloadRaw;
+    return (
+      <Address address={payload.address}>
+        <Bytes bytes={payload.data} url={url} />
+        {isFirst && !isExternal && <Unlock onSign={_onSign} />}
+        <ActionBar>
+          <div />
+          <Link isDanger onClick={_onCancel}>Cancel</Link>
+        </ActionBar>
+      </Address>
+    );
+  } else {
+    return null;
+  }
 }
--- a/packages/extension/src/background/RequestBytesSign.ts
+++ b/packages/extension/src/background/RequestBytesSign.ts
@@ -1,0 +1,23 @@
+// Copyright 2019 @polkadot/extension authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import { KeyringPair } from '@polkadot/keyring/types';
+import { RequestSign } from './types';
+import { SignerPayloadJSON, SignerPayloadRaw } from '@polkadot/types/types';
+import { u8aToHex, hexToU8a } from '@polkadot/util';
+import { TypeRegistry } from '@polkadot/types';
+
+export default class RequestBytesSign implements RequestSign {
+  inner: SignerPayloadJSON | SignerPayloadRaw;
+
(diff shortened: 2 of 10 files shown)


jacogr:
Thanks for doing this, this is a great addition. I will take a look in the next couple of hours.

packages/extension-ui/src/Popup/Signing/Request.tsx
```diff
@@ -3,37 +3,43 @@
...
 import { createType } from '@polkadot/types';
 
 import { ActionBar, ActionContext, Address, Link } from '../../components';
 import { approveSignPassword, approveSignSignature, cancelSignRequest } from '../../messaging';
-import Details from './Details';
+import Extrinsic from './Extrinsic';
+import Bytes from './Bytes';
 import Qr from './Qr';
 import Unlock from './Unlock';
+import { SignerPayloadJSON, SignerPayloadRaw } from '@polkadot/types/types';
```
jacogr: (small nitpick) Move this to the "types section", i.e. the first bundle of imports - alphabetical, so will end up on line 7 with line 8 being the new blank.

packages/extension/src/background/types.ts
```diff
@@ -63,7 +64,8 @@ export interface RequestSignatures {
   'pub(accounts.list)': [RequestAccountList, InjectedAccount[]];
   'pub(accounts.subscribe)': [RequestAccountSubscribe, boolean, InjectedAccount[]];
   'pub(authorize.tab)': [RequestAuthorizeTab, null];
-  'pub(extrinsic.sign)': [RequestExtrinsicSign, ResponseExtrinsicSign];
+  'pub(extrinsic.sign)': [SignerPayloadJSON, ResponseSigning];
+  'pub(bytes.sign)': [SignerPayloadRaw, ResponseSigning];
```
jacogr: (nitcpick) Alphabetical, so just one up.

packages/extension/src/background/handlers/Tabs.ts
```diff
@@ -4,11 +4,15 @@
...
 import { InjectedAccount } from '@polkadot/extension-inject/types';
 import { SubjectInfo } from '@polkadot/ui-keyring/observable/types';
-import { RequestAuthorizeTab, RequestExtrinsicSign, ResponseExtrinsicSign, RequestTypes, ResponseTypes, MessageTypes } from '../types';
+import { RequestAuthorizeTab, ResponseSigning, RequestTypes, ResponseTypes, MessageTypes } from '../types';
+import { KeyringPair } from '@polkadot/keyring/types';
 
 import keyring from '@polkadot/ui-keyring';
 import accountsObservable from '@polkadot/ui-keyring/observable/accounts';
 import { assert } from '@polkadot/util';
+import { SignerPayloadJSON, SignerPayloadRaw } from '@polkadot/types/types';
```
jacogr: (nitpick) Move to types.

packages/extension-ui/src/Popup/Signing/Request.tsx
```diff
@@ -3,37 +3,43 @@
...
 
 export default function Request ({ account: { isExternal }, isFirst, request, signId, url }: Props): React.ReactElement<Props> | null {
   const onAction = useContext(ActionContext);
-  const [payload, setPayload] = useState<ExtrinsicPayload | null>(null);
+  const [hexBytes, setHexBytes] = useState<string | null>(null);
+  const [extrinsic, setExtrinsic] = useState<ExtrinsicPayload | null>(null);
 
   useEffect((): void => {
-    setPayload(createType('ExtrinsicPayload', request, { version: request.version }));
+    const inner = request.getInner();
```
jacogr: Probably would have implemented this as an enum-like structure, where you can do `isRaw` & `getRaw` (and `isPayload` & `getPayload`, but this is fine, it works as expected.

c410-f3r:
Code updated. Thanks for reviewing

c410-f3r:
@jacogr I tested the implementation manually and fixed some issues

jacogr:
Thanks for that, it is the only thing I was still missing to merge - I have it top of the list for the morning. (I am truly sorry for the long delay between the PR and merge here, not quite cool seeing thing stuck, so my humble apologies)

jacogr:
Via this one, https://github.com/polkadot-js/apps/pull/1936, the signature comes back as bytes instead of hex -

image

It should be hex.

c410-f3r:
Updated

jacogr:
Thank you. Will do the final play-through this morning, really excited to get this in.

jacogr:
A couple of issues -

1. When I sign text, e.g. `hello` it works. When I sign hex data e.g. `0x123456` the signature does not match. I believe it is due to the fact that we pass the data through the TextEncoder - we should just pass the raw input data

2. When I try and send an extrinsic, `pub(extrinsic.sign)` passed, it still brings up the bytes signing dialog.This is a screenshot of the 2, the first being the current extrinsic, the second being actual bytes

image

3. The text styling for `from` & `bytes` above does not match up with the actual old extinsic display and looks out of place. Compare the above against this -

image

jacogr:
I have merged support for this on the apps UI, so this feature should be easier to test and explore now.

c410-f3r:
WTF! So such problems. Sorry for the noise.
The code has been updated to reflect the issues

jacogr:
This one is tough to test properly without having some way of well, testing :) So no reason to apologize, I also looked at the code and well, that looks good - however, you were playing in the depths of midnight, so those excursions are never easy.

packages/extension/src/background/RequestBytesSign.ts
```diff
@@ -0,0 +1,22 @@
...
+export default class RequestBytesSign implements RequestSign {
+  inner: SignerPayloadJSON | SignerPayloadRaw;
+
+  constructor (inner: SignerPayloadRaw) {
+    this.inner = inner;
+  }
+
+  sign (pair: KeyringPair): { signature: string } {
+    const inner = this.inner as SignerPayloadRaw;
+    const signedBytes = pair.sign(stringToU8a(inner.data));
```
c410-f3r: According to the `SignerPayloadRawBase` signature, 

```typescript
export interface SignerPayloadRawBase {
    /**
     * @description The hex-encoded data for this request
     */
    data: string;
    /**
     * @description The type of the contained data
     */
    type?: 'bytes' | 'payload';
}
```

`data` will always be hex, so it is up to the caller to pass valid data.

Also, `type?: 'bytes' | 'payload';` doesn't make sense. One can simply `const bytes = extrinsic.toU8a(true)` and `signRaw(bytes)`. Can we remove it?

c410-f3r:
> This one is tough to test properly without having some way of well, testing :) So no reason to apologize, I also looked at the code and well, that looks good - however, you were playing in the depths of midnight, so those excursions are never easy.

Indeed! These are good examples of a programming life

packages/extension/src/background/RequestBytesSign.ts
```diff
@@ -0,0 +1,22 @@
...
+export default class RequestBytesSign implements RequestSign {
+  inner: SignerPayloadJSON | SignerPayloadRaw;
+
+  constructor (inner: SignerPayloadRaw) {
+    this.inner = inner;
+  }
+
+  sign (pair: KeyringPair): { signature: string } {
+    const inner = this.inner as SignerPayloadRaw;
+    const signedBytes = pair.sign(hexToU8a(inner.data));
```
c410-f3r: According to the `SignerPayloadRawBase` signature, 

```typescript
export interface SignerPayloadRawBase {
    /**
     * @description The hex-encoded data for this request
     */
    data: string;
    /**
     * @description The type of the contained data
     */
    type?: 'bytes' | 'payload';
}
```

`data` will always be hex, so it is up to the caller to pass valid data.

Also, `type?: 'bytes' | 'payload';` doesn't make sense. One can simply `const bytes = extrinsic.toU8a(true)` and `signRaw(bytes)`. Can we remove it?

jacogr: The `payload` was added specifically for other extensions, such as the centrality one - where they don't care about de-serliazing the payload, but rather just want to sign it as a raw blob. From an implementation perspective, both can just be treated as `bytes` here, it does not need to be treated separately. 

(The reason behind it - in case of transactions, the API will first try to use the normal payload sign, if that is not available will fallback to `signRaw`. Since the normal sign path is catered for, we should not receive a payload. 

Either way, it it ever does receive a `type: payload` it actually cannot quite decode it, since it is in a signable form, not a decodable form, so it should treat it as a "piece of data". (There is a small tick here which came in with v4 where the signature needs to have type information attached in the case of a `payload`, but we don't need to cater for it here)

TL;DR Ignore the type specifier on `signRaw`, we can just treat is all as bytes.

c410-f3r: Very interesting, thank you for writing this in an explanatory way

jacogr:
I'll need to take a peek again at the code here, since something is going wrong, either in the extension or somewhere along the way. When submitting a transaction (verified against current, it doesn't happen), the transactions come through as "immortal", not "mortal with era", while the API def submits mortals.

image

jacogr:
So the issue with transactions is twofold -

- the request is wrongly displayed, i.e. the mortal information is missing, which indicated the decoding was not properly done
- the generated signature fails, i.e. the transaction cannot be submitted

(Both these are verified on an older version as working on the same chain)

I believe is is due to how inner is constructed. So rather,

- the message that is sent, instead of doing an "inner" with a specific type, just keep it as-is for the data received, this will always be hex. We don't need to convert via type to anything, when signing we can just . keep the logic as-is. This means -

- for bytes, pass it as e are currently doing here
- for the normal extrinsic sign, keep this in-place https://github.com/polkadot-js/extension/pull/196/files#diff-dcc23ab59637c1964f03628674b307f8L163 

With the above change reverted, at least the signing portion should still work, I'm trying to figure . out wht the `.inner` and cast makes the decoding not work in . the way it was.

c410-f3r:
My recent sickness plus some other responsibilities temporally took my motivation to work on this PR.

Well, back on track. Here is a screenshot of a correctly displayed substrate transaction info.

Capture d’écran de 2019-11-29 20-23-41

And another showing the same transaction successfully registered.

Capture d’écran de 2019-11-29 20-25-02

jacogr:
Thank you !

Hope you are doing ok.

jacogr:
Sorry, I'm very aware I need a play-through and review here ASAP. Just "a bit" snowed under after the Kusama CC3 launch. So need to, want to, I just have not had a gap.

My humble apologies for dragging my feet, not out of choice.

c410-f3r:
@jacogr 

Don't worry about it, we all have stuff to do. Personally, you are one of the most attentive reviewers I have ever met

jacogr:
I will take a play through today, thanks for the patience on this :(

c410-f3r:
This is awesome! Thanks!

c410-f3r:
@jacogr Do you know when will be next marketplace deployment?

jacogr:
I will do it soon - I am just stuck with a nasty flu atm, so don’t quite trust a release process. So just need an update to latest Kusama CC3 and a full test through. 

I am hoping in the next couple of days (hopefully latest on Monday), I just need a clear head. ;)

c410-f3r:
Flu catches us all. Hope you get better
Thanks for the information!

PR #253, opened 2020-02-25
Provider Injection

amaurymartiny:
reviving #88

- Added 5 messages: `pub(rpc.send)`, `pub(rpc.subscribe)`, `pub(rpc.listProviders)`, `pub(rpc.startProvider)`, `pub(rpc.unsubscribe)`

cc @sjeohp

### DIFF
--- a/packages/extension-dapp/README.md
+++ b/packages/extension-dapp/README.md
@@ -13,7 +13,7 @@ A basic extractor that manipulates the `window.injectedWeb3` to retrieve all the
 ## Usage
 
 ```js
-import { web3Accounts, web3Enable, web3FromAddress, web3Providers } from '@polkadot/extension-dapp';
+import { web3Accounts, web3Enable, web3FromAddress, web3RpcProvider } from '@polkadot/extension-dapp';
 
 // returns an array of all the injected sources
 // (this needs to be called first, before other requests)
@@ -39,5 +39,5 @@ api.tx.balances
 // retrieve a RPC provider from a particular source
 // if the extension does not expose this provider, it will return an error.
 const provider = web3RpcProvider('polkadot-js', { type: 'WsProvider', payload: 'http://locahost:9944' });
-const head = await provider.send('chain_getHeader'); // logs latest header
+const head = await provider.send('chain_getHeader', []); // shows latest header
 ```
--- a/packages/extension-dapp/README.md
+++ b/packages/extension-dapp/README.md
@@ -36,8 +36,11 @@ api.tx.balances
   .transfer('5C5555yEXUcmEJ5kkcCMvdZjUo7NGJiQJMS7vZXEeoMhj3VQ', 123456)
   .signAndSend('5DTestUPts3kjeXSTMyerHihn1uwMfLj8vU8sqF7qYrFabHE', (status) => { ... });
 
-// retrieve a RPC provider from a particular source
-// if the extension does not expose this provider, it will return an error.
-const provider = web3RpcProvider('polkadot-js', { type: 'WsProvider', payload: 'http://locahost:9944' });
+// retrieve all the RPC providers from a particular source. this returns a map
+// of string->ProviderMeta, where the string represents an unique key to identify
+// a provider, and ProviderMeta shows metadata about this provider
+const allProviders = web3ListRpcProviders('polkadot-js');
+// assuming one of the keys in `allProviders` is 'kusama-cc3', we can then use that provider
+const { provider } = web3UseRpcProvider('polkadot-js', 'kusama-cc3');
 const head = await provider.send('chain_getHeader', []); // shows latest header
 ```
--- a/packages/extension-dapp/README.md
+++ b/packages/extension-dapp/README.md
@@ -13,7 +13,7 @@ A basic extractor that manipulates the `window.injectedWeb3` to retrieve all the
 ## Usage
 
 ```js
-import { web3Accounts, web3Enable, web3FromAddress } from '@polkadot/extension-dapp';
+import { web3Accounts, web3Enable, web3FromAddress, web3Providers } from '@polkadot/extension-dapp';
 
 // returns an array of all the injected sources
 // (this needs to be called first, before other requests)
@@ -35,4 +35,9 @@ api.setSigner(injector.signer);
 api.tx.balances
   .transfer('5C5555yEXUcmEJ5kkcCMvdZjUo7NGJiQJMS7vZXEeoMhj3VQ', 123456)
   .signAndSend('5DTestUPts3kjeXSTMyerHihn1uwMfLj8vU8sqF7qYrFabHE', (status) => { ... });
+
+// retrieve a RPC provider from a particular source
+// if the extension does not expose this provider, it will return an error.
+const provider = web3RpcProvider('polkadot-js', { type: 'WsProvider', payload: 'http://locahost:9944' });
+const head = await provider.send('chain_getHeader'); // logs latest header
 ```
--- a/packages/extension-dapp/README.md
+++ b/packages/extension-dapp/README.md
@@ -13,7 +13,7 @@ A basic extractor that manipulates the `window.injectedWeb3` to retrieve all the
 ## Usage
 
 ```js
-import { web3Accounts, web3Enable, web3FromAddress, web3RpcProvider } from '@polkadot/extension-dapp';
+import { web3Accounts, web3Enable, web3FromAddress, web3ListRpcProviders, web3UseRpcProvider } from '@polkadot/extension-dapp';
 
 // returns an array of all the injected sources
 // (this needs to be called first, before other requests)
--- a/packages/extension-dapp/src/index.ts
+++ b/packages/extension-dapp/src/index.ts
@@ -2,7 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import { Injected, InjectedAccount, InjectedAccountWithMeta, InjectedExtension, InjectedExtensionInfo, InjectedProviderWithMeta, InjectedWindow, Unsubcall } from '@polkadot/extension-inject/types';
+import { Injected, InjectedAccount, InjectedAccountWithMeta, InjectedExtension, InjectedExtensionInfo, InjectedProviderWithMeta, InjectedWindow, ProviderJSON, Unsubcall } from '@polkadot/extension-inject/types';
 
 // just a helper (otherwise we cast all-over, so shorter and more readable)
 const win = window as Window & InjectedWindow;
@@ -178,21 +178,30 @@ export async function web3FromAddress (address: string): Promise<InjectedExtensi
 }
 
 // retrieve all the providers accross all extensions
-export async function web3Providers (): Promise<InjectedProviderWithMeta[]> {
+export async function web3Providers (providerJSON: ProviderJSON): Promise<InjectedProviderWithMeta[]> {
   if (!web3EnablePromise) {
     return throwError('web3Providers');
   }
 
   const injected = await web3EnablePromise;
-  const providers: InjectedProviderWithMeta[] = injected
-    .map(({ provider, name: source }): InjectedProviderWithMeta | null =>
-      provider
-        ? { provider, meta: { source } }
-        : null
-    )
-    .filter((x): x is InjectedProviderWithMeta =>
-      x !== null
-    );
+
+  const providers: InjectedProviderWithMeta[] = (await Promise.all(
+    injected
+      .map(async ({ provider, name: source }): Promise<InjectedProviderWithMeta | null> => {
+        try {
+          if (!provider) {
+            return null;
+          }
+
+          await provider.setProvider(providerJSON);
+
+          return { provider, meta: { ...providerJSON, source } };
+        } catch {
+          return null;
+        }
+      })
+
+  )).filter((x): x is InjectedProviderWithMeta => x !== null);
 
   const sources = providers.map(({ meta: { source } }): string => source);
 
--- a/packages/extension-dapp/src/index.ts
+++ b/packages/extension-dapp/src/index.ts
@@ -2,7 +2,7 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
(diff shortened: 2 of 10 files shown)


packages/extension-dapp/README.md
```diff
@@ -35,4 +35,9 @@ api.setSigner(injector.signer);
 api.tx.balances
   .transfer('5C5555yEXUcmEJ5kkcCMvdZjUo7NGJiQJMS7vZXEeoMhj3VQ', 123456)
   .signAndSend('5DTestUPts3kjeXSTMyerHihn1uwMfLj8vU8sqF7qYrFabHE', (status) => { ... });
+
+// retrieve a RPC provider from a particular source
+// if the extension does not expose this provider, it will return an error.
+const provider = web3RpcProvider('polkadot-js', { type: 'WsProvider', payload: 'http://locahost:9944' });
```
amaurymartiny: I went with a blacklist method: user can request any provider from extension (for now: only WsProvider, any endpoint), and background will allow/reject the request.

An alternative would be a whitelist method:
- user sends `pub(rpc.getAllProviders)`, the background will return all the Providers it can provide
- user sends `pub(rpc.startProvider)`, the background instantiates that provider (aka call `new WhateverProvider()`)

packages/extension/src/background/handlers/State.ts
```diff
@@ -213,6 +222,40 @@ export default class State {
...
+  public rpcSubscribe (request: RequestRpcSubscribe, cb: ProviderInterfaceCallback): Promise<number> {
+    assert(this.#provider, 'Cannot call pub(rpc.subscribe) before provider has been set');
+
+    return this.#provider.subscribe(
+      request.type,
+      request.method,
+      request.params,
+      // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
+      // @ts-ignore WsProvider gives me (error, result)=>void, whereas (result)=>void is expected
+      (_error, res) => { cb(res); }
```
amaurymartiny: this is weird... ProviderInterface's `subscribe` method has signature `(result)=>void`, but somehow with WsProvider I get `error` (null most of the time) as 1st arg

axelchalon: ah, yes, I remember bumping into this issue. the signature in WsProvider is wrong. you're right, [`subscribe`](https://github.com/polkadot-js/api/blob/333ff8528b790eb790e5d31f951b01ed35c3c746/packages/rpc-provider/src/ws/Provider.ts#L236) takes a [`ProviderInterfaceCallback`](https://github.com/polkadot-js/api/blob/333ff8528b790eb790e5d31f951b01ed35c3c746/packages/rpc-provider/src/types.ts#L39) of signature `(result)=>void` but the callback is actually called [this way](https://github.com/polkadot-js/api/blob/333ff8528b790eb790e5d31f951b01ed35c3c746/packages/rpc-provider/src/ws/Provider.ts#L352-L357), i.e. with `(_error, res)`. I guess the ProviderInterfaceCallback signature should be changed

packages/extension/src/background/handlers/State.ts
```diff
@@ -213,6 +222,40 @@ export default class State {
...
+    assert(this.#provider, 'Cannot call pub(rpc.subscribe) before provider has been set');
+
+    return this.#provider.send(request.method, request.params);
+  }
+
+  // Change the provider the extension is exposing
+  public rpcSetProvider (provider: RequestRpcProvider): Promise<null> {
+    switch (provider.type) {
+      case 'WsProvider': {
+        this.#provider = new WsProvider(provider.payload);
```
axelchalon: haven't tested, but looking at the code it seems me to me that

- when the page closes, the wsprovider connections it has made should be closed, which doesn't seem to be the case here (subscriptions are left active), and
- unexpected behaviour if two pages use the extension (i.e. if two pages are opened which use `rpcSetProvider`, the latter will create a new WsProvider, and the existing connections from the former page will use the previous WsProvider which doesn't have any more references to it, so memory leaks, while the next subscriptions from the former page will use the new WsProvider -- also any page can change the extension provider endpoint used by all the other pages?). in my PR I had solved this by mapping a page (chrome.runtime.Port) to a provider ([code](https://github.com/polkadot-js/extension/pull/88/files#diff-08c560d3145c89a48bc10bc086ce5fe0R239)) and using one WsProvider connection per page

packages/extension/src/background/handlers/Tabs.ts
```diff
@@ -72,6 +73,28 @@ export default class Tabs {
     return this.#state.sign(url, new RequestExtrinsicSign(request), { address, ...pair.meta });
   }
 
+  private rpcSend (request: RequestRpcSend): Promise<JsonRpcResponse> {
+    return this.#state.rpcSend(request);
```
axelchalon: if the page sends a request and the user closes the page before the response arrives, the background script is going to throw (line 34)

https://github.com/polkadot-js/extension/blob/602dfdc0262b62e142bfa47163b93b262a184d49/packages/extension/src/background/handlers/index.ts#L26-L40

packages/extension-dapp/README.md
axelchalon: just a thought, but in the future maybe we should restrict the use of one `web3RpcProvider` (`setProvider`) per page, otherwise a page we allowed might DoS an endpoint if it keeps calling `setProvider` to the same endpoint (or variations thereof) to connect/disconnect to it repeatedly

packages/extension/src/background/handlers/Tabs.ts
```diff
@@ -72,6 +73,28 @@ export default class Tabs {
...
+  }
+
+  private rpcSetProvider (request: RequestRpcProvider): Promise<null> {
+    return this.#state.rpcSetProvider(request);
+  }
+
+  private async rpcSubscribe (request: RequestRpcSubscribe, id: string, port: chrome.runtime.Port): Promise<boolean> {
+    const cb = createSubscription<'pub(accounts.subscribe)'>(id, port);
```
axelchalon: just a thought, `id` is provided by the webpage (which can call `postMessage` manually, bypassing `extension-inject`), so a malicious page could provide the `id` of an existing subscription made by another page (iirc it's incremental so wouldn't be difficult) and it would overwrite the mapping in `subscriptions`. this would mess up our logic of unsubscribing to a page's subscriptions when the page closes.

packages/extension/src/background/types.ts
```diff
@@ -216,5 +232,5 @@ export type MessageTypesWithNoSubscriptions = Exclude<MessageTypes, keyof Subscr
 export interface RequestSign {
   readonly inner: SignerPayloadJSON | SignerPayloadRaw;
 
-  sign (registry: TypeRegistry, pair: KeyringPair): { signature: string };
+  sign(registry: TypeRegistry, pair: KeyringPair): { signature: string };
```
jacogr:
```suggestion
  sign (registry: TypeRegistry, pair: KeyringPair): { signature: string };
```

packages/extension/src/page/PostMessageProvider.ts
```diff
@@ -0,0 +1,153 @@
...
+        this.#subscriptions[`${type}::${id}`] = callback;
+
+        return id;
+      });
+    } else {
+      return this.#sendRequest('pub(rpc.send)', { method, params });
+    }
+  }
+
+  public async setProvider (providerJSON: ProviderJSON): Promise<null> {
```
jacogr: Remove. The provider injected should be pre-configured.

packages/extension-dapp/README.md
jacogr: The idea is that when the injected page already contains a pre-configured provider as set. So effectively they can, without doing anything else, just use the providers.

So the interface should just provide a list of all providers, all of them already configured as per the specific extensions. So some additional metadata can be provided -

- source
- name of provider source, e.g. W3F or Parity or ...
- network, e.g Kusama or Custom...
- type (full or light)

packages/extension-dapp/src/index.ts
```diff
@@ -176,3 +176,22 @@ export async function web3FromAddress (address: string): Promise<InjectedExtensi
 
   return web3FromSource(found.meta.source);
 }
+
+// retrieve a provider from one source
+export async function web3RpcProvider (source: string, providerJSON: ProviderJSON): Promise<InjectedProviderWithMeta | null> {
```
jacogr: Just source here, no JSON or setting. (It should not be driven by the user, rather the extension to set all info)

packages/extension-dapp/README.md
amaurymartiny: right, I went with the pre-configured providers way

amaurymartiny:
Addressed some of the comments.

I personally still think that the extension may expose to dapps multiple Providers.

So now the dapp user would do:
```typescript
      const injected = await web3Enable()
      const allProviders = await web3ListRpcProviders(injected[0].name)
      console.log(allProviders) // Map of ProviderKey -> ProviderMeta

      // Assuming one key of `allProviders` is kusama-cc3-ws
      const { provider } = await web3UseRpcProvider(injected[0].name, "kusama-cc3-ws")

      const api = new Api({ provider })
```

packages/extension-inject/src/types.ts
```diff
@@ -3,6 +3,7 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import { Signer as InjectedSigner } from '@polkadot/api/types';
+import PostMessageProvider from '@polkadot/extension/page/PostMessageProvider';
```
amaurymartiny: Circular dep

packages/extension/src/background/handlers/State.ts
```diff
@@ -59,7 +68,11 @@ export default class State {
 
   readonly #authRequests: Record<string, AuthRequest> = {};
 
-  #provider: ProviderInterface | undefined;
+  // Map of providers currently injected in tabs
+  readonly #injectedProviders: Map<chrome.runtime.Port, ProviderInterface> = new Map()
```
jacogr:
```suggestion
  readonly #injectedProviders: Map<chrome.runtime.Port, ProviderInterface> = new Map();
```

Weird that the linter didn't complain.

packages/extension/src/background/handlers/State.ts
```diff
@@ -222,31 +244,48 @@ export default class State {
...
-    return Promise.resolve(null);
+    assert(Object.keys(this.#providers).includes(key), `Provider ${key} is not exposed by extension`);
+
+    // Instantiate the provider
+    this.#injectedProviders.set(port, this.#providers[key].start());
+
+    // Close provider connection when page is closed
+    port.onDisconnect.addListener((): void => {
+      const provider = this.#injectedProviders.get(port);
+      if (provider) { provider.disconnect(); }
```
jacogr:
```suggestion
      if (provider) {
        provider.disconnect();
      }
```

packages/extension/src/background/handlers/State.ts
```diff
@@ -256,6 +295,13 @@ export default class State {
     );
   }
 
+  public rpcUnsubscribe (request: RequestRpcUnsubscribe, port: chrome.runtime.Port): Promise<boolean> {
+    const provider = this.getProvider(port);
+    assert(provider, 'Cannot call pub(rpc.unsubscribe) before provider has been set');
```
jacogr:
```suggestion

    assert(provider, 'Cannot call pub(rpc.unsubscribe) before provider has been set');
```

packages/extension/src/background/handlers/State.ts
```diff
@@ -222,31 +244,48 @@ export default class State {
...
+    });
+
+    return Promise.resolve(this.#providers[key].meta);
   }
 
-  public rpcSubscribe (request: RequestRpcSubscribe, cb: ProviderInterfaceCallback): Promise<number> {
-    assert(this.#provider, 'Cannot call pub(rpc.subscribe) before provider has been set');
+  public rpcSubscribe (request: RequestRpcSubscribe, cb: ProviderInterfaceCallback, port: chrome.runtime.Port): Promise<number> {
+    const provider = this.getProvider(port);
+    assert(provider, 'Cannot call pub(rpc.subscribe) before provider has been set');
```
jacogr:
```suggestion

    assert(provider, 'Cannot call pub(rpc.subscribe) before provider has been set');
```

packages/extension-dapp/README.md
jacogr: Docs need an update here.

packages/extension-dapp/src/index.ts
```diff
@@ -176,3 +176,28 @@ export async function web3FromAddress (address: string): Promise<InjectedExtensi
 
   return web3FromSource(found.meta.source);
 }
+
+// retrieve all providers exposed by one source
+export async function web3ListRpcProviders (source: string): Promise<ProviderList | null> {
+  const { provider } = await web3FromSource(source);
+  if (!provider) {
```
jacogr:
```suggestion

  if (!provider) {
```

packages/extension-inject/src/types.ts
jacogr: Can just create an interface, i.e. `InjectedProvider`, the provider can implement it. (Slight extra work, but probably the more correct approach)

amaurymartiny:
Ok, tested  by adding a WsProvider at this line: https://github.com/polkadot-js/extension/blob/master/packages/extension/src/background/handlers/index.ts#L12, it works.

Only https://github.com/polkadot-js/extension/pull/253#discussion_r384168174 needs to addressed, i'll have a quick look if there's an easy way to do it

jacogr:
"I personally still think that the extension may expose to dapps multiple Providers."

We can change the interface to return `providers?: Provider[]`

amaurymartiny:
Well, for the dapp user, the interface will always be a PostMessageProvider, so my reasoning was that one was enough. Then this PostMessageProvider selects the actual background provider to use (with `.startProvider(key)`).

The alternative would be to spawn one PostMessageProvider per background provider. Could work too, in which as it'd have the signature you proposed. But then it would uselessly spawn `n-1` other PostMessageProviders.

jacogr:
Fair enough. I'm happy to have a `source` & `'key` for multiples. 

As long as the key is provided by the extension itself, all good). It is like the `Signer`, the key you provide tells it where to act. And a list of these keys are provided by the accounts.

packages/extension/src/background/handlers/Tabs.ts
amaurymartiny: proposing #254

PR #279, opened 2020-03-19
PostMessageProvider with on('connected' | 'disconnected')

amaurymartiny:
Let PostMessageProvider know if the upstream background provider is connected or not


packages/extension/src/background/types.ts
```diff
@@ -240,5 +241,5 @@ export type MessageTypesWithNoSubscriptions = Exclude<MessageTypes, keyof Subscr
 export interface RequestSign {
   readonly inner: SignerPayloadJSON | SignerPayloadRaw;
 
-  sign (registry: TypeRegistry, pair: KeyringPair): { signature: string };
+  sign(registry: TypeRegistry, pair: KeyringPair): { signature: string };
```
amaurymartiny:
```suggestion
  sign (registry: TypeRegistry, pair: KeyringPair): { signature: string };
```

jacogr:
Seems sane, just needs a master merge. (After `-base`)

PR #292, opened 2020-03-26
Fix @types references (sadly, def dupe)

jacogr:


packages/extension/src/extensionizer.d.ts
```diff
@@ -0,0 +1,11 @@
+// Copyright 2019-2020 @polkadot/extension authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+/// <reference types="chrome"/>
+
+declare module 'extensionizer' {
+  const extensionizer: typeof chrome;
```
amaurymartiny: afaiu, extensionizer is a wrapper to let`chrome` behave like WebExtension. But `chrome` itself doesn't follow WebExtension.

The correct way would be `const extensionizer: browser;` here, where `browser` comes from WebExtensions (maybe no need for additional types pkg, or else from https://github.com/kelseasy/web-ext-types)

PR #297, opened 2020-03-30
Publish `extension-chains`

amaurymartiny:
extension-base, published to npm, needs extension-chains:

https://github.com/polkadot-js/extension/blob/6d838335453d29537d3fbf2e0c5d36c298567252/packages/extension-base/src/background/handlers/State.ts#L1-L5


jacogr:
Have not merged yet, since this one actually needs to go the way of the dodo. So atm on  the fence to publish and then whack, so not 100% sure on the future here as of yet. (And need the current stable version available first to see the actual "update metadata" impact on users here)

jacogr:
Done in #315 alongside other dep cleanups.

PR #299, opened 2020-03-31
Add validation to text inputs

vanruch:


packages/extension-ui/src/components/ValidatedInput.tsx
```diff
@@ -0,0 +1,57 @@
+// Copyright 2019 @polkadot/extension-ui authors & contributors
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+import React, { useState, useEffect } from 'react';
+
+import styled from 'styled-components';
+import { Result, Validator } from '@polkadot/extension-ui/validators';
```
jacogr: Could just be `../validators`

packages/extension-ui/src/components/ValidatedInput.tsx
```diff
@@ -0,0 +1,57 @@
...
+      setValidationResult(result);
+      onChange(Result.isOk(result) ? value : null);
+    })();
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [value, validator, onChange]);
+
+  return (
+    <div className={className}>
+      {React.cloneElement(children, { isError: Result.isError(validationResult), value, onChange: setValue })}
+      {Result.isError(validationResult) && <ErrorMessage>{validationResult.error.errorDescription}</ErrorMessage>}
```
jacogr: Is there no other way of doing this? (The clone & isError React items)

packages/extension-ui/src/partials/DerivationPath.tsx
```diff
@@ -3,8 +3,9 @@
 // of the Apache-2.0 license. See the LICENSE file for details.
 
 import React, { useCallback, useState } from 'react';
-import { InputWithLabel } from '@polkadot/extension-ui/components';
+import { InputWithLabel, ValidatedInput } from '@polkadot/extension-ui/components';
```
jacogr: Just single imports from '../components/<name>

packages/extension-ui/src/partials/DerivationPath.tsx
```diff
@@ -3,8 +3,9 @@
...
 import { validateDerivationPath } from '@polkadot/extension-ui/messaging';
+import { Result } from '@polkadot/extension-ui/validators';
```
jacogr: ../validators

packages/extension-ui/src/partials/Name.tsx
```diff
@@ -2,50 +2,41 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import React, { useContext, useEffect, useState } from 'react';
+import React, { useContext } from 'react';
 
-import { AccountContext, InputWithLabel } from '../components';
+import { AccountContext, InputWithLabel, ValidatedInput } from '../components';
+import { isNotShorterThan } from '@polkadot/extension-ui/validators';
```
jacogr: ../validators

packages/extension-ui/src/partials/Password.tsx
```diff
@@ -2,47 +2,55 @@
 // This software may be modified and distributed under the terms
 // of the Apache-2.0 license. See the LICENSE file for details.
 
-import React, { useEffect, useState } from 'react';
+import React, { useState, useEffect } from 'react';
 
-import { InputWithLabel } from '../components';
+import { InputWithLabel, ValidatedInput } from '../components';
+import { allOf, isNotShorterThan, isSameAs, Validator } from '@polkadot/extension-ui/validators';
```
jacogr: ../validators

packages/extension-ui/src/components/ValidatedInput.tsx
vanruch: We could pass callback with error, value and callback. Actually tried doing it that way at first, but it was much uglier and less readable

jacogr: Really uncomfortable with this approach, especially around `clone` i.e. next stable release of React recommends `React.jsx(...)` as opposed to `React.createElement(...)`, so dunno what will happen going forward.

There could always be a hook that can create all these callbacks and then passed as-is (so 1 value instead of 3 of them - so the user has a single line to take care of and the magic happens here)

vanruch: Removed `cloneElement`

jacogr:
Ok, it looks good, I just broke linting ... again. (Last time today though, just wanted to get all the little things into all the repos)

PR #306, opened 2020-04-02
Show labels of account parent

vanruch:


packages/extension-base/src/background/types.ts
```diff
@@ -23,11 +23,22 @@ type NullKeys<T> = { [K in keyof T]: IsNull<T, K> }[keyof T];
...
 export type SeedLengths = 12 | 24;
 
-export interface AccountJson {
+export interface AccountJson extends KeyringPair$Meta {
   address: string;
   genesisHash?: string | null;
   isExternal?: boolean;
   name?: string;
+  parentAddress?: string;
+  derivationPath?: string;
```
jacogr, line 32: Not convinced we want to store this at all - it could also contain a password `///<Password>` which mean that when exported, that is all in the clear.

packages/extension-ui/src/utils/buildHierarchy.test.ts
```diff
@@ -0,0 +1,55 @@
+import { AccountJson, AccountWithChildren } from '@polkadot/extension-base/background/types';
```
jacogr, line 1: Just missing the header here.

packages/extension-ui/src/utils/buildHierarchy.ts
```diff
@@ -0,0 +1,23 @@
+import { AccountJson, AccountWithChildren } from '@polkadot/extension-base/background/types';
```
jacogr, line 1: Header missing.

packages/extension-ui/src/utils/buildHierarchy.ts
```diff
@@ -0,0 +1,23 @@
...
+
+  return {
+    ...account,
+    children
+  };
+};
+
+export const buildHierarchy = (accounts: AccountJson[]): AccountWithChildren[] => accounts
+  .filter(isRoot)
+  .map(accountWithChildren(accounts));
```
jacogr, line 23:
```suggestion
export const buildHierarchy = (accounts: AccountJson[]): AccountWithChildren[] =>
  accounts
    .filter(isRoot)
    .map(accountWithChildren(accounts));
```

packages/extension-base/src/background/types.ts
vanruch: We can't include ///<Password> into derivation path for some reason. I am getting `Error: Re-constructed path "//path//password" does not match input` error from keyring when trying to do so

packages/extension-ui/src/utils/buildHierarchy.ts
vanruch: Maybe consider adding this plugin to linter?
https://www.npmjs.com/package/eslint-plugin-header

jacogr: Will add it. I hate playing manual linter. (And I certainly also miss a lot of stuff when I code and then have to shout at myself down the line)

packages/extension-base/src/background/types.ts
jacogr: That is weird. Will look into that, weirdly I did the same on the apps UI day or so ago and it was fine of account -> account derivation. Don't worry about it here, will play and try to track it down to not waste time.

vanruch: I thought, it was intended to be like that :). I'll remove derivationPath from metadata then.

packages/extension-base/src/background/handlers/Extension.test.ts
```diff
@@ -107,6 +107,7 @@ describe('Extension', () => {
         suri: '//path'
       }, {} as chrome.runtime.Port);
       expect(keyring.getAccount('5FP3TT3EruYBNh8YM8yoxsreMx7uZv1J1zNX7fFhoC5enwmN')?.meta.parentAddress).toEqual(address);
+      expect(keyring.getAccount('5FP3TT3EruYBNh8YM8yoxsreMx7uZv1J1zNX7fFhoC5enwmN')?.meta.derivationPath).toEqual('//path');
```
jacogr, line 110: Just whack it :)

```suggestion
```

PR #309, opened 2020-04-03
Remove extensionizer indirection

jacogr:


packages/extension-base/src/extension.ts
```diff
@@ -0,0 +1,11 @@
...
+// This software may be modified and distributed under the terms
+// of the Apache-2.0 license. See the LICENSE file for details.
+
+const extension = typeof chrome !== 'undefined'
+  ? chrome
+  : typeof browser !== 'undefined'
+    ? browser
+    : null;
+
+export default extension as typeof chrome;
```
amaurymartiny, line 11: Why not:
- Use this "official" polyfill from Mozilla: https://github.com/mozilla/webextension-polyfill
- Use `browser.*.*` everywhere (as per WebExtension standard)
- (optionally) Add TS with https://github.com/kelseasy/web-ext-types#install-it

This will make the extension WebExtension-compatible.

Currently it seems like jumping through hoops imo.

jacogr: This is all driven by actual Chrome store approvals - which is a huge issue due to the indirection via extensionizer here.

So, point taken, but the answer is a resounding "NO".

There are other issues at play here.

Issue #310, opened 2020-04-03
Remove persistent background pages (basically eol in Chrome)


jacogr:
Specifically https://github.com/polkadot-js/extension/blob/master/packages/extension/manifest.json#L11

This puts the extension down the road to get ready for a move to V3 manifests (persistent not supported, background pages are service workers)

There is an impact on initialization, so will see what the overhead turns out to be. (Specifically wasm-crypto)

This also means that the in-memory state needs to be moved to actual storage as applicable.

The checklist -

https://developer.chrome.com/docs/extensions/mv3/mv3-migration-checklist/

jak-pan:
Ah so this is why in chrome canary the extension breaks after using it for a while. Stable is now at version 89 where it works. Canary is at 91 so there's probably still some time.

jacogr:
Umpf. There is a still a lot of work to do on this front... not ready for v3 as of yet :(

jak-pan:
It looks like almost all of the extension is working even without the wasm and with the error present. It's just the popup window account screen that doesn't work. Even account creation from the popup is ok. Isn't there just some easy fix for now, that would at least partly make it seem not broken? I mean it's usable...

jacogr:
I saw the UI issue on canary, it is the css calc here that is broken -

![image](https://user-images.githubusercontent.com/1424473/112051877-baa6ba00-8b52-11eb-99c2-3119c0cd7bc2.png)

It really looks like a Canary bug. Setting it to `height: 100vh` works - but you cannot quite do it each and every time you open it...

jak-pan:
Ah sorry. I thought that it has to do something with the error in the console. 

Maybe OT but:
I rememeber exactly this problem happening several years back, when working on an extension. It had to do with the window starting as zero size and resizing to content, but it didn't always fire repaints correctly, so we needed to hack it, by introducing resize after load. I thought I'm crazy by telling this but I still found this is an issue from time to time until today! https://stackoverflow.com/questions/8840580/force-dom-redraw-refresh-on-chrome-mac

jacogr:
:) Just spend some time reading a number of interesting workarounds - and then got to stuff like „October 2020 - the above hacks stopped working, here are some new ones“

zeedgold:
This is very interesting

zeedgold:
This is very interesting

Nick-1979:
Are there any plans to work on this serious issue in the near future?

jacogr:
[...]

PR #312, opened 2020-04-03
Introduce base store (with key prefix mgmt)

jacogr:
Make store generic for extended use, e.g. https://github.com/polkadot-js/extension/issues/310


jacogr:
Really should move the ui-keyring ExtensionStore over as well, but that does multiple keys. So not yet. Someday.

PR #318, opened 2020-04-17
Create new account as a child of a master account by default

vanruch:


jacogr:
So couple of comments as I played-through -

- "root account" sounds very scary, maybe "parent account" rather? (Especially when I de-selected, it said "new root", which made me wonder what happens to my other "root". Here is would actually probably e useful to rather say (now that I typed it out) "from a new seed"

- on creating a child, the path is hidden. I would suggest just removing the toggle and always displaying the derivation path. We could also make this rather a toggle, so by default it is "non-editable", but really should be visible.

- when I create another root, I have no way of actually using it to derive from. The first one is the default. So here we can probably have a drop-down of all the available parent accounts. (Well, actually it is all accounts, since you can derive from a derived account... but probably best to keep to parent only, derivation we can still do per-account)

- feels a bit complex, when toggle/untoggle, may be better with the first point. I just slightly feel like I am losing track of what I am doing. Maybe the headers adjust at least on the initial screen when you can tick.

Overall happy with the general idea, just feels "more difficult to do things now", not always 100% sure as an idiot user what the intent is.

krzysztof-jelski:
> * when I create another root, I have no way of actually using it to derive from. The first one is the default. So here we can probably have a drop-down of all the available parent accounts. (Well, actually it is all accounts, since you can derive from a derived account... but probably best to keep to parent only, derivation we can still do per-account)

There is a dropdown -- when there are 2 roots, you can click the account under "Nest account under".

jacogr:
Ok, _now_ that you mention it, I see it... not sure how the I missed that at all. Just looked like the normal card :(

jacogr:
I am still on this, thanks for the patience, just a small roadblock to a prompt review on my side.

PR #336, opened 2020-05-22
Restore Account with JSON Backup

shawntabrizi:
This is an early draft of enabling the Polkadot.JS extension to restore accounts using the JSON backup created by the Polkadot JS Apps.

TODO:

- [x] Positive/Negative feedback in UI
    - [x] When file selected is good or bad
    - [x] When password is correct or incorrect
    - [x] When restore is successful or not
- [x] Styling
- [x] "upload" SVG icon


shawntabrizi:
I seem to have got it working:

image

A few things here:

* What does `keyring.isPassValid` do when not doing a full restore? It does not take in the JSON file, so it just makes sure the string provided has certain requirements?

* How do I handle the async messages between the extension and browser? i.e. I want to update a `useState` or show the user a message as a result of sending a message across the border. Do I just need to add `async`/`await`?

shawntabrizi:
Updated Styling:

1. 
image

2. 
image

3. 
image

4. 
image

5. 
image

6. 
image

shawntabrizi:
@jacogr can you help me fix the linting issues?

I am unable to get the lint error to appear locally, and tbh not quite sure how to solve all of them correctly.

jacogr:
Yes, will help fix, no issues. (With the recent bump to typescript-eslint 3 there are actually some linter bugs, so happy to help fix.) 

Will get to this shortly, really excited to have it - things have just been a bit hectic elsewhere, so this hasn't gotten the attention it deserves.

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -177,6 +177,51 @@ export default class Extension {
...
+    try {
+      const pair = keyring.restoreAccount(json, password);
+
+      if (pair) {
+        const { address } = pair;
+
+        return { message: `Successfully added ${address}` };
+      }
+    } catch (error) {
+      return { message: `${error}` };
```
jacogr, line 190:
```suggestion
      return { message: (error as Error).message };
```

(lint fix. the issue is that in JS a `throw` can be anything, even a string, but since it is only our stuff, we know these are always an `Error`)

packages/extension-ui/src/Popup/RestoreJson.tsx
```diff
@@ -0,0 +1,108 @@
...
+import React, { useCallback, useState } from 'react';
+import { InputWithLabel, InputFileWithLabel, Button, Address } from '../components';
+import { u8aToString } from '@polkadot/util';
+import styled from 'styled-components';
+import { jsonRestore, jsonVerifyPassword, jsonVerifyFile } from '../messaging';
+
+import { Header } from '../partials';
+
+type Props = {};
+
```
jacogr, line 15:
```suggestion
```

Remove for linting, issue below...

packages/extension-ui/src/Popup/RestoreJson.tsx
```diff
@@ -0,0 +1,108 @@
...
+    return { address, isFileValid, json };
+  } catch (error) {
+    console.error(error);
+  }
+
+  return { address: null, isFileValid: false, json: null };
+}
+
+export default function Upload (): React.ReactElement<Props> {
```
jacogr, line 43:
```suggestion
export default function Upload (): React.ReactElement {
```

lint fix, in conjunction with the issue above. (once again, typescript-eslint got stricter, this used to be ok)

packages/extension-ui/src/Popup/RestoreJson.tsx
```diff
@@ -0,0 +1,108 @@
...
+interface PassState {
+  isPassValid: boolean;
+  password: string;
+}
+
+const acceptedFormats = ['application/json', 'text/plain'].join(', ');
+
+async function parseFile (file: Uint8Array): Promise<FileState> {
+  try {
+    const json = JSON.parse(u8aToString(file));
```
jacogr, line 31:
```suggestion
    const json = JSON.parse(u8aToString(file)) as KeyringPair$Json;
```

We need to explicitly cast, since the parsing output is untyped. (Basically saying, he, this is what I expect)

jacogr: (Also fixes the linting errors below, since the type is now known)

packages/extension-ui/src/Popup/RestoreJson.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+  return (
+    <>
+      <HeaderWithSmallerMargin
+        text='Restore Account from JSON'
+      />
+      <div>
+        <Address
+          address={isFileValid && address ? address : null}
+          name={isFileValid && json ? json.meta.name : null}
```
jacogr, line 79:
```suggestion
          name={isFileValid && json ? json.meta.name as string : null}
```

The meta is typed as `unknown` for all fields, here we just way what we believe it to be. (Still in two minds on this actually, since we can actually tighten this up significantly since there is actually known fields, it is not quite "just a holder of anything")

jacogr:
PS: Once again, sorry it took me absolute ages to get to this, not making excuses, but it was a "bit tight" lately.

shawntabrizi:
Awesome! Can't wait to use this feature :)

PR #393, opened 2020-07-21
Updates for JSON v3

### LINKED ISSUE #391: Support for kdf JSON
As per https://github.com/polkadot-js/common/pull/637

We would need a small progress/spinner while unlocking as well, it is nowhere near instaneous anymore.

### LINKED ISSUE #392: Press return key to sign tx on the popup
Currently when the popup shows and I entered the password, I have to click "Sing this transaction" button. Jut hitting the return key doesn't work. It could be very nice if we can add this.

### LINKED ISSUE #394: Password length
Its not allowing me to import json when setting up an account if the password is say 40 characters.  the password length allowed should be upped.

jacogr:
- Closes #391
- Closes #392 
- Closes #394


packages/extension-ui/src/Popup/Export.tsx
```diff
@@ -91,6 +98,10 @@ const MovedWarning = styled(Warning)`
 
 const ActionArea = styled.div`
   padding: 10px 24px;
+
+  > .spinButton {
+    position: relative;
+  }
```
jacogr, line 104:
```suggestion
```

PR #408, opened 2020-08-19
add crab network

WoeOm:
website:
https://darwinia.network/

telemetry:
https://telemetry.polkadot.io/#/Crab

explorer:
https://crab.subscan.io/block/0

### DIFF
--- a/packages/extension-chains/src/chains.ts
+++ b/packages/extension-chains/src/chains.ts
@@ -30,6 +30,12 @@ const hashes: MetadataDefBase[] = [
     genesisHash: '0xf7a99d3cb92853d00d5275c971c132c074636256583fee53b3bbe60d7b8769ba',
     icon: 'substrate',
     ss58Format: 16
+  },
+  {
+    chain: 'Crab',
+    genesisHash: '0x34f61bfda344b3fad3c3e38832a91448b3c613b199eb23e5110a635d71c13c65',
+    icon: 'substrate',
+    ss58Format: 42
   }
 ];
 
--- a/packages/extension-chains/src/chains.ts
+++ b/packages/extension-chains/src/chains.ts
@@ -20,6 +20,12 @@ const hashes: MetadataDefBase[] = [
   },
   // alphabetical to avoid (further) favorites
   {
+    chain: 'Crab',
+    genesisHash: '0x34f61bfda344b3fad3c3e38832a91448b3c613b199eb23e5110a635d71c13c65',
+    icon: 'substrate',
+    ss58Format: 42
+  },
+  {
     chain: 'Edgeware',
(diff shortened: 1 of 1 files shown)


packages/extension-chains/src/chains.ts
```diff
@@ -30,6 +30,12 @@ const hashes: MetadataDefBase[] = [
     genesisHash: '0xf7a99d3cb92853d00d5275c971c132c074636256583fee53b3bbe60d7b8769ba',
     icon: 'substrate',
     ss58Format: 16
+  },
+  {
```
jacogr, line 34: Can you just move this up to right under "alphabetical to avoid (further) favorites" - the alphabetical ordering makes it easier to find/maintain as the list grows.

jacogr:
As a live network, is there a reason it used the `42` (Substrate default/development) prefix? Would suggest registering it in https://github.com/paritytech/substrate/blob/master/primitives/core/src/crypto.rs#L414

packages/extension-chains/src/chains.ts
WoeOm: done

WoeOm:
> As a live network, is there a reason it used the `42` (Substrate default/development) prefix? Would suggest registering it in https://github.com/paritytech/substrate/blob/master/primitives/core/src/crypto.rs#L414

Crab is a canary network，there is no specific prefix to the network

AurevoirXavier:
> As a live network, is there a reason it used the `42` (Substrate default/development) prefix? Would suggest registering it in https://github.com/paritytech/substrate/blob/master/primitives/core/src/crypto.rs#L414

It's darwinia's testnet. Like kusama and polkadot's relation. And we don't want to waste a prefix. So it's 42 : )

jacogr:
The reason for asking is that I've tried to not add all the testnets in the extension and keep it to value-bearing chains. Which is why things like Flaming Fir, WestEnd, Rocco, etc are not in here for the Polkadot-related chains. 

Unlike things like the apps UI to deploy any changes is very slow (mostly driven by store approvals which can take 10 working days on publish - we do get luckier with a "couple of" working day turnarounds normally.)

So don't want to waste your time -

- really happy with live/canary value-bearing chains
- not quite excited about testnets since they add maintenance overhead

So don't want to throw any work here out the window, but just want to make sure we are aligned on the above and this is indeed value-bearing before opening the door for a flood of non-value testnets.

AurevoirXavier:
Actually, this's a valuable chain. Like I mentioned before. 

> Like kusama and polkadot's relation.

Later we'll bridge the crab to darwinia. The tokens can be transfer to the mainnet with ratio 1:100.

jacogr (approved):

AurevoirXavier:
Thanks!

jacogr:
Always a pleasure to merge any external PRs. 

Sorry for the interrogation, but I'm sure you understand where my questions come from :)

Issue #274, opened 2020-03-17
Sign a message or tx silently w/o asking for a password for X period of time.


siman:
There should be an option to sign a message or tx silently w/o asking for a password for X period of time. At least this could be implemented per browser tab session: tab open -> session started -> ask for a password once -> /* ... don't ask a password for other txs during this session ... */ -> tab closed -> session stoped.

## Use case

A user has unread notifications with a counter on Subsocial. When a user goes to the "Notifications" page, we need to send a query to backend (with Postgres) to mark all notifications as read and set unread count to `0`. Much like on Twitter.

This means that we need to sign a message on behalf of a current account that he/she wants to mark all unread notifications as read.

Showing extension modal and asking a current user to sign a message just when the user goes to the "Notifications" page doesn't sound like a good UX at all.

Issue #275, opened 2020-03-17
Don't ask confirmation for a certain set of extrinsics per chain


siman:
## Use case

At Subsocial there are social extrinsics: Comment a post, Upvote/Downvote a post or comment.
This should be relatively cheap or free txs and it would be a good UX to not bother users with a confirmation modal for such kind of txs.

## Proposal

If there was a way to specify a set of extrinsics per chain that doesn't require a confirmation - this would be cool and would help UX of the dapps in general!

## UPD

So if there was a way for a dapp to tell through the extension to a user:
Would you like that we will not ask you a confirmation for these set of extrinsics?
[ ] palletX.extrinsicA
[ ] palletY.extrinsicB
...
[ ] palletZ.extrinsicC
[ ] Yes to all

jacogr:
Will have to think through this - no asking the user to unlock (after initial) as per https://github.com/polkadot-js/extension/issues/274 (message, app, chain) is really good. However doing things without the use knowing is probably not a direction we just want to stumble into.

So in #274 we can greatly smooth the process, but not convinced we want to take control away.

siman:
Added this to the description:

So if there was a way for a dapp to tell through the extension to a user:
Would you like that we will not ask you a confirmation for these set of extrinsics?
[ ] palletX.extrinsicA
[ ] palletY.extrinsicB
...
[ ] palletZ.extrinsicC
[ ] Yes to all

siman:
These settings can be stored per browser + per account.
Dapp is proposing a set of `[pallet + extrinsi]+` to the extension.
And if a user click OK or some checkboxes, ext. just remembers the choices in local storage?

siman:
@Tbaut added my recent thoughts from our conversation

jacogr:
I am not open to this at all - have done this is an Ethereum app, and it was unusable. The complexity is too high for the user to make any sense and right calls. 

So this won’t go on.

jacogr:
This issue has been open for 21 days with no activity and is not labelled as an enhancement. It will be closed in 7 days.

siman:
Works be happy to contribute this discussion. This is very needed
functionality. Just ask Subsocial users. Not every blockchain's UX is
transferring tokens once per week. People hate seeing these confirmation
modals even with a checked "Remember me" box

Best regards,
Alex Siman

jacogr:
[...]

PR #421, opened 2020-09-03
Don't ask for the password for 15 minutes

jacogr:
Closes https://github.com/polkadot-js/extension/issues/274

### DIFF
--- a/packages/extension-ui/src/Popup/Signing/Request.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Request.tsx
@@ -43,10 +43,11 @@ export default function Request ({ account: { isExternal }, buttonText, isFirst,
   const [{ hexBytes, payload }, setData] = useState<Data>({ hexBytes: null, payload: null });
   const [error, setError] = useState<string | null>(null);
   const [isBusy, setIsBusy] = useState(false);
-  const [isLocked, setIsLocked] = useState(true);
+  const [isLocked, setIsLocked] = useState<boolean | null>(null);
   const [isSavedPass, setIsSavedPass] = useState(false);
 
   useEffect((): void => {
+    setIsLocked(null);
     !isExternal && isSignLocked(signId)
       .then((isLocked) => setIsLocked(isLocked))
       .catch((error: Error) => console.error(error));
@@ -81,7 +82,7 @@ export default function Request ({ account: { isExternal }, buttonText, isFirst,
     (password: string): Promise<void> => {
       setIsBusy(true);
 
-      return approveSignPassword(signId, password, !!password && isSavedPass)
+      return approveSignPassword(signId, password, !!(password && isSavedPass))
         .then((): void => {
           setIsBusy(false);
           onAction();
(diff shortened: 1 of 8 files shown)


packages/extension-ui/src/Popup/Signing/Request.tsx
```diff
@@ -61,24 +71,70 @@ export default function Request ({ account: { isExternal }, buttonText, isFirst,
...
+    ? (
+      <Unlock
+        buttonText={buttonText}
+        error={error}
+        isBusy={isBusy}
+        onSign={_onSign}
+      >
+        <Checkbox
+          checked={!!isSavedPass}
+          label="Don't ask me again for the next 15 minutes"
```
Tbaut, line 125: Not familiar with the code of the extension (yet). Would it be possible to have this `15` not hard-coded without introducing too much complexity?

Tbaut (approved):
Working well on Firefox!

packages/extension-ui/src/Popup/Signing/Request.tsx
jacogr: The extension is as minimal as possible (and way too complex, just don't get time to fix...). Happy to consider it if people explicitly do ask for it. Since it means -

(a) extending settings to allow you to enter
(b) storing it with the settings
(c) making it available to both UI & background

Initially thought about a dropdown for 1m, 5m, 15m, 30m & 60m, which is less complex than the above, but way too many clicks again. (Although it could default to some sane-ish)

Tbaut: Ah sorry, this is not what I meant.
I meant that this "15" itself is duplicated and harcoded in 2 different placed, although it would have been more generic to export a variable `const DELAY = 15` next to [PASSWORD_EXPIRY](https://github.com/polkadot-js/extension/pull/421/files#diff-037fe702f2f94ba52ce14b32737d9c74R24) and use this "DELAY" in the label.

I'm totally with you on the 15 as a good starting point, a dropdown or anythibg else would be too heavy visually.

siman:
This is really a good one for UX of the dapps! ❤️

packages/extension-ui/src/Popup/Signing/Request.tsx
siman: I vote for multiple options. If you don't want dropdowns, then what about chips? Like:
[ 15 mins ] [ 30 mins ] [ 60 mins ]

jacogr:
Sorry it took so long... not having this functionality is actually driving me up the wall currently as an active user...

siman:
Another UX must-have feature is: Don't ask confirmation for this list of dispatchable methods in a form of `${moduleName}.${txName}`. A dapp on connection request could ask if a user is ok if they will not be asked for a password for this list of tx methods.

jacogr:
See #275 (& #274) - at this point, not convinced on either. Both add a lot of complexity (and have done exactly his before with whitelisting on Ethereum. I have not closed either, so it is not off the table, but it is not something that will be taken lightly since it add a large maintenance and UX burden.)

Anyway, this PR is close, so done and dusted, any conversation of related functionality needs to go into their respective issues, cannot keep on following dead merged PRs.

Issue #409, opened 2020-08-20
doc/help: bundle setup


thelostone-mc:
#### Description 

I'm trying to use browserify to generate a bundle using browserify that I can add to my django app (for Gitcoin)

When I try to generate the bundle 
https://github.com/thelostone-mc/polkadot-bundle/blob/master/ext-dependencies.js

I get this 
![image](https://user-images.githubusercontent.com/5358146/90806843-36315600-e33b-11ea-82ad-0f1af81f5792.png)

I'm assuming I'd need both `extension` / `extension-ui` to be able to trigger a transfer transaction from the extension right ?
Not sure if I'm missing anything obvi

The bundle gets created when I use only `extension-dapp`
but then since the underlying connections / UI packages aren't there -> I end up with empty array

![image](https://user-images.githubusercontent.com/5358146/90807583-4f86d200-e33c-11ea-98ad-e889e15d0748.png)

thelostone-mc:
The goal is to be able to send KSM tokens from A (signed in via extension) -> B  
and the same for DOT token

jacogr:
You should only need extension-dapp to allow the extension to work in your app.

thelostone-mc:
@jacogr Ah okay then bundle generation works but I end up with this
Would I have to add enable else ?

https://user-images.githubusercontent.com/5358146/90807583-4f86d200-e33c-11ea-98ad-e889e15d0748.png

jacogr:
I believe you probably need to pass a name for your app to web3Enable as per https://github.com/polkadot-js/extension/tree/master/packages/extension-dapp#usage

That should popup a dialog asking you to approve.

thelostone-mc:
Oh perfect ! That works 
Btw this extension is pretty dope <3

thelostone-mc:
Oh perfect ! That works 
Btw this extension is pretty dope <3

thelostone-mc:
Oh perfect ! That works 
Btw this extension is pretty dope <3

jacogr:
Phew! That was a shot in the dark... I'm really happy my guess helped.

jacogr:
... it really should do _something_ if you pass `undefined` there though, even a helpful error would help.

So re-opening this to just track that as an enhancement so nobody else steps in the same hole.

thelostone-mc:
@jacogr I'll keep iterating on this meanwhile so in case if anyone wants to use it -> they could port over the bundle and roll with it 
[...]

PR #430, opened 2020-09-04
Throw error for undefined name in web3Enable

Tbaut:
linked to https://github.com/polkadot-js/extension/issues/409
Is that totally useless @jacogr ?


jacogr (approved):
Not useless. People have stepped into this, so the runtime check makes sense.

PR #438, opened 2020-09-07
Fix back buttonnot appearing

Tbaut:
fixing the following where the back arrow in not showing on the button.
image


jacogr (approved):
Thank you.

PR #463, opened 2020-09-21
Share PASSWORD_EXPIRY default between components

Tbaut:
Just a couple nits found as I'm getting more familiar with the extension's code.


jacogr:
Please update this as well - https://github.com/polkadot-js/extension/blob/master/packages/extension/public/locales/zh/translation.json#L89

Just the default input string without tweaking zh will work. (But can `s/15/{{minutes}}/`)

Tbaut:
Thanks, I updated the english (although it's probably automatic?) and the zh one.

packages/extension/public/locales/zh/translation.json
```diff
@@ -86,7 +86,7 @@
   "Sign the transaction": "签名此交易",
   "Sign the message": "签名此消息",
   "Scan signature via camera": "通过摄像头扫描签名",
-  "Don't ask me again for the next 15 minutes": "在接下来的15分钟内别再问我了",
+  "Don't ask me again for the next {{expiration}} minutes": "在接下来的{{expiration}}分钟内别再问我了",
```
jacogr, line 89: Thank you! Really don't want this to go out of style before we have it in as a selectable language. (Not sure if that is logged, but should really just start of simple, no detect and allow swapping. This is our biggest non-English market by far)

jacogr: PS: Yes, English _should_ get done at CI time. (But the unknown keys would be dropped as part of this, so in this case we preserve the translations)

jacogr (approved):
Thank you!

Issue #400, opened 2020-07-31
UI closes when switching tabs or window


marceljay:
On Firefox 79, macOS Catalina:

When creating a new account from a new seed, the moment you switch tabs to paste the seed phrase, the UI closes and you have to restart the process. 
A user not paying attention could even restart the process, not noticing a new seed phrase, thinking they already copied it and thus creating a new account which they have **not** stored the seed phrase for. 

Proposed solution: When switching back to the window or tab, and opening the extension from the button, the last state should appear again, at least for some time (e.g. 240 seconds) to save the seed in e.g. a password manager.

Tbaut:
Regarding the implementation, I'm thinking about context variable that is set to `{mnemonic, ttl: now() + 2min}` that would allow an automatic redirection. The one part that bothers me is the TTL reset. As long as there is no interaction, the variable with the mnemonic will stay in memory, potentially longer than 2min. We could eventually take advantage of [`chrome.alarm`](https://developer.chrome.com/extensions/alarms) to make sure the variable is effectively reset. What do you think?

jacogr:
It is fine if it is active in the background page, it is not accessible. (So the timeout needs to be managed there)

jak-pan:
Easy solution that other wallets use is to open a new browser tab when doing this. You don't lose state here. 

The proposed solution is probably not ideal, because:
1. you will be sending and storing mnemonic
2. when the timer expires which is probably not seen by the user, it will be a problem anyway. (Unless you explicitly show timer for the account). 

The question then is with this solution, how long do you want that timer to be? Since user might want to also write it down and can close the popup while reading a guide for example (this "bug" occured to some users when they were claiming and reading a guide).

I can also see a possibility of unwanted mnemonic storage persistence. Don't see direct problem on the first sight, but it should be considered.

jacogr:
Indeed, the opening the account in a tab is imho, an acceptable workaround and one that is available to all users. Not perfect, but doesn't close on the user.

Longer-term would love to move all the vault views/management to only operate in a tab (lots more real-estate to play wih), but it has not had any focus.

luyenheo:
@`
![unnamed](https://user-images.githubusercontent.com/115646489/225676909-a2c16227-7929-4d3e-867f-3fa0c629fce4.jpg)
`

ap211unitech:
This issue appears to be outdated. However, feel free to open a new issue with updated details if you're still experiencing the problem.

PR #466, opened 2020-09-25
Redirect to account creation when aborted

Tbaut:
closes https://github.com/polkadot-js/extension/issues/400


packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -21,6 +21,7 @@ type CachedUnlocks = Record<string, number>;
 
 const SEED_DEFAULT_LENGTH = 12;
 const SEED_LENGTHS = [12, 15, 18, 21, 24];
+const CACHE_ACCOUNT_TIMEOUT = 10 * 1000;
```
Tbaut, line 24:
```suggestion
const CACHE_ACCOUNT_TIMEOUT = 2 * 60 * 1000;
```

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -178,6 +181,25 @@ export default class Extension {
     return true;
   }
 
+  private flushAccountCache (): void {
+    console.log('flushing account cache..');
```
Tbaut, line 185:
```suggestion
```

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -178,6 +181,25 @@ export default class Extension {
...
+    this.cachedAccount = undefined;
+  }
+
+  private setAccountCache (account: AutoSavedAccount): void {
+    console.log('setting cached account to', account);
```
Tbaut, line 190:
```suggestion
```

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -178,6 +181,25 @@ export default class Extension {
...
+    this.cachedAccount = account;
+
+    setTimeout(() => {
+      console.log('timeout kicking in');
```
Tbaut, line 194:
```suggestion
```

packages/extension-ui/src/Popup/index.tsx
```diff
@@ -103,14 +108,27 @@ export default function Popup (): React.ReactElement {
       .catch(console.error);
   }, [cameraOn]);
 
+  useEffect(() => {
+    getAccountCache()
+      .then((cachedAccount) => {
+        if (cachedAccount?.seed) {
+          setHasCachedAccount(true);
+          // _onAction('/account/create');
```
Tbaut, line 116:
```suggestion
```

packages/extension-ui/src/partials/HeaderWithSteps.tsx
```diff
@@ -11,16 +11,20 @@ import Header from './Header';
...
   className?: string;
+  onCancel?: () => void;
   step: number;
   text: string;
 }
 
-function HeaderWithSteps ({ className, step, text }: Props): React.ReactElement<Props> {
+function HeaderWithSteps ({ className, onCancel, step, text }: Props): React.ReactElement<Props> {
   const onAction = useContext(ActionContext);
+  // const { setAutoSavedAccount } = useContext(AutoSavedAccountRedirectContext);
```
Tbaut, line 21:
```suggestion);
```

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -178,6 +181,22 @@ export default class Extension {
     return true;
   }
 
+  private flushAccountCache (): void {
+    this.cachedAccount = undefined;
```
jacogr, line 185: On flush you need to cancel the timeout. Otherwise use may create 1, start the time, start a new one and othe original timer clears.

packages/extension-ui/src/Popup/Accounts/index.tsx
```diff
@@ -4,11 +4,11 @@
 import React, { useContext } from 'react';
 import styled from 'styled-components';
 
+import AccountsTree from './AccountsTree';
+import AddAccount from './AddAccount';
```
jacogr, line 8: local directory imports come last, so it was correct to follow after partials.

packages/extension-ui/src/Popup/CreateAccount/index.tsx
```diff
@@ -3,9 +3,10 @@
 
 import React, { useCallback, useContext, useEffect, useState } from 'react';
 
+import { createAccountSuri, createSeed, flushAccountCache, getAccountCache, setAccountCache } from '@polkadot/extension-ui/messaging';
```
jacogr, line 6: just ../messaging, we are in the current package.

packages/extension-ui/src/Popup/CreateAccount/index.tsx
```diff
@@ -18,9 +19,20 @@ export default function CreateAccount (): React.ReactElement {
...
+      .then((cachedAccount) => {
+        if (cachedAccount?.address) {
+          setAccount(cachedAccount);
+        } else {
+          createSeed()
+            .then((account) => {
+              setAccount(account);
+              setAccountCache(account).catch((e) => console.error(e));
+            })
+            .catch((error: Error) => console.error(error));
```
jacogr, line 32: Can remove this .catch, return the promise. (The outer catch will handle it)

packages/extension-ui/src/Popup/CreateAccount/index.tsx
```diff
@@ -43,10 +58,16 @@ export default function CreateAccount (): React.ReactElement {
 
   const _onNextStep = useCallback(() => setStep((step) => step + 1), []);
   const _onPreviousStep = useCallback(() => setStep((step) => step - 1), []);
+  const _onCancel = useCallback(() => {
+    flushAccountCache().catch((e) => console.error(e));
+    onAction('/', { resetCachedAccount: true });
```
jacogr, line 63: Ummm... see what you mean. Not sure about this one at all. Will need to locally look at the logic and step through, but this seems strange.

packages/extension-ui/src/Popup/CreateAccount/index.tsx
Tbaut: @typescript-eslint/no-floating-promises is complaining then (that's pretty new afaik):
> Promises must be handled appropriately or explicitly marked as ignored with the `void` operator.

Adding the void makes it complain for another reason
> Expected 'undefined' and instead saw 'void'.

packages/extension-base/src/background/handlers/Extension.ts
Tbaut: Thanks I added it. I've set the timeout in the state then, that looks a little wild though, wondering if I should rather declare those variables in the `./State`?

packages/extension-ui/src/Popup/CreateAccount/index.tsx
jacogr: Weird. Then it would need to stay just to satisfy the linter. All ok.

(I'm not too fond of the "void" solution, the `.catch(console.error)` at least tells me I'm not setting anything, but at least am reporting on it.

packages/extension-base/src/background/handlers/Extension.ts
jacogr: Will take a look. But my knee-jerk was that it is messy to manage, since you need to track the timerId as well. Don't see another solution though.

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -37,6 +38,9 @@ export default class Extension {
 
   readonly #state: State;
 
+  private cachedAccount: AutoSavedAccount | undefined;
+  private cacheTimeOut: number | undefined;
```
jacogr, line 42:
```suggestion
  #cachedAccount: AutoSavedAccount | undefined;
  #cacheTimeOut: number | undefined;
```

Rather use Js privates since TS supports it.

packages/extension-ui/src/Popup/CreateAccount/CreateAccount.test.tsx
```diff
@@ -185,4 +190,26 @@ describe('Create Account', () => {
       expect(wrapper.find('[data-button-action="add new root"] button').prop('disabled')).toBe(false);
     });
   });
+
+  describe('Account gets cached and restored', () => {
```
jacogr, line 194: +1 (and another +1 and ...) for adding UI tests. I suck at these.

packages/extension-ui/src/Popup/CreateAccount/index.tsx
```diff
@@ -43,10 +57,16 @@ export default function CreateAccount (): React.ReactElement {
...
```
jacogr, line 62: If you are passing the reset flag, why are we explicitly flushing the cache here and above? While the flag feels a bit messy (it is growing on me), just making it explicit and using it everywhere is the better/cleaner/more-maintainable option

packages/extension-ui/src/Popup/CreateAccount/CreateAccount.test.tsx
Tbaut: trying to get better at this, but there's still a long way.
I'm only partly satisfied by these tests because:
- if I test [here](https://github.com/polkadot-js/extension/pull/466/files#diff-1f14364b73a398aa6bfdcd7926ad6c99R211) for
```diff
- expect(getAccountCache).toBeCalled();
+ expect(getAccountCache).toBeCalledTimes(1);
```

it fails because it's called twice.. but I don't understand why.

Tbaut:
Does this PR have a chance to get merged or what do you think it would need so that it's merged? We don't have to have this somewhat hacky part as param in the `onAction`, it causes a little flash of the UI because there's a redirection, but that's very subtle.

packages/extension-base/src/background/types.ts
```diff
@@ -350,4 +363,9 @@ export interface ResponseJsonRestore {
   error: string | null;
 }
 
+export interface ActionOptions {
+  resetCachedAccount?: boolean;
+}
+
 export type AllowedPath = typeof ALLOWED_PATH[number];
+
```
Tbaut, line 371:
```suggestion
```

packages/extension-ui/src/Popup/CreateAccount/index.tsx
Tbaut: I got rid of this, opting for an "AccountCache" context.

Tbaut:
closing this one until we have figured out better ways to solve this problem.

PR #467, opened 2020-09-28
Add i18n language change dropdown

### LINKED ISSUE #465: Allow i18n language selection
We now have english & Chinese. Just another dropdown in the settings dialog that switches. (The apps UI allows switching, so can check there)

Tbaut:
closes #465


Tbaut:
I've needed quite some copy/paste from apps (`i18n/Backend`, `i18n/cache`, `util/languages.ts`). lkm if this is something we should share in ui-setting somewhere.

packages/extension-ui/src/util/languages.ts
```diff
@@ -0,0 +1,30 @@
...
+  isDisabled?: boolean;
+  isHeader?: boolean;
+  text: React.ReactNode;
+  value: string | number;
+}
+
+export default function create (t: TFunction): Option[] {
+  return [
+    {
+      text: t<string>('lng.detect', 'Default browser language (auto-detect)', { ns: 'apps-config' }),
```
Tbaut, line 17:
```suggestion
      text: t<string>('lng.detect', 'Browser language (auto-detect)'),
```

packages/extension-ui/src/util/languages.ts
```diff
@@ -0,0 +1,30 @@
+// Copyright 2017-2020 @polkadot/apps-config authors & contributors
```
Tbaut, line 1:
```suggestion
 // Copyright 2019-2020 @polkadot/extension-ui authors & contributors
```

jacogr:
Was actually thinking of not using detection here at all. Just the languages English/Chinese. This also simplified the backend quite a bit.

packages/extension-ui/src/util/languages.ts
jacogr: Probably (if going for auto-detection) it would make sense to just remove the "(auto-detect)" rather.

jacogr (commented):
Seems sane. (Even without my no-detect suggestion)

packages/extension-ui/src/util/languages.ts
Tbaut: Sounds good, I removed the language detection.

Tbaut:
:+1: I ditched the language detection.

> This also simplified the backend quite a bit.

You did not mean the backend that is in the `Backend` file right? Although what happens there is over my head, I understand that this tells where to get the translations, which seems to be independent from the language detection.

jacogr:
The backend stuff there is overly complex, but that is due to apps where changing translations need to reflect while doing. Hence all the caching. That ended up being try #5 or something, there are still gaps in the interaction with the translations as done, i.e. live refresh, but it does _something_.

jacogr (approved):
Thank you very much!

PR #471, opened 2020-09-30
Add error boundaries with fallback component to show any error

### LINKED ISSUE #469: Add ErrorBoundary (as applicable)
Wile https://github.com/polkadot-js/extension/issues/468 (based on description) seems like a background crash (???), the UI itself should never go "white page" or drop out. ErrorBoundary (as used in the apps UI) protects against that.

Need at least one for the authorization/signing where things may go wrong due to bad parsing.

Tbaut:
closes #469


packages/extension-ui/src/Popup/index.tsx
```diff
@@ -114,39 +116,41 @@ export default function Popup (): React.ReactElement {
...
-                    </ToastProvider>
-                  </SigningReqContext.Provider>
-                </MetadataReqContext.Provider>
-              </MediaContext.Provider>
-            </AuthorizeReqContext.Provider>
-          </AccountContext.Provider>
-        </SettingsContext.Provider>
-      </ActionContext.Provider>
-    )}</Loading>
+    <ErrorBoundary FallbackComponent={ErrorFallback}>
```
jacogr, line 119: The issue with only having it this far up the tree is that you can never recover. So it would need to sit inside the signing component and still have a "cancel" when it gets into this state.

However... probably need a reproduction to see it working. (Well, I guess we can make it crash on purpose in the affected parts)

yarn.lock
```diff
@@ -3077,6 +3077,7 @@ __metadata:
     react: ^16.13.1
     react-dom: ^16.13.1
     react-dropzone: ^11.1.0
+    react-error-boundary: ^3.0.2
```
jacogr, line 3080: Would rather prefer not pulling in a dep here. It is not that complex since React has it as a base - https://github.com/polkadot-js/apps/blob/master/packages/react-components/src/ErrorBoundary.tsx

Tbaut: Sounds good, I thought it's simple, but maintained by high profile devs so that should have covered us pretty well, but having our own class is also fine, I didn't know we had it in apps.

packages/extension-ui/src/Popup/index.tsx
Tbaut: Right, having it this high in the tree was a first attempt so try to see more helpful errors.
I'll play with it lower.

jacogr: 100%, the first round is just catching... hopefully :)

packages/extension-ui/src/Popup/Authorize/index.tsx
```diff
@@ -14,9 +14,9 @@ export default function Authorize (): React.ReactElement {
   const requests = useContext(AuthorizeReqContext);
 
   return (
-    <>
-      <Scroll isLastRequest={requests.length === 1}>
-        <Header text={t<string>('Authorize')} />
+    <Scroll isLastRequest={requests.length === 1}>
```
jacogr, line 17: Just worried the removal of the fragment here breaks stuff - there are a couple `> div` which was hugely problematic styling-wise in the past.

packages/extension-ui/src/Popup/index.tsx
```diff
@@ -114,39 +114,41 @@ export default function Popup (): React.ReactElement {
...
-                    </ToastProvider>
-                  </SigningReqContext.Provider>
-                </MetadataReqContext.Provider>
-              </MediaContext.Provider>
-            </AuthorizeReqContext.Provider>
-          </AccountContext.Provider>
-        </SettingsContext.Provider>
-      </ActionContext.Provider>
-    )}</Loading>
+    <ErrorBoundary>
```
Tbaut, line 117: Does this make sense to keep this here still in case some unexpected error isn't caught by any `ErrorBoundary` sitting below?

packages/extension-ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+import React from 'react';
+
+import translate from './translate';
+
+interface Props extends WithTranslation {
+  children: React.ReactNode;
+  doThrow?: boolean;
+  error?: Error | null;
+  onError?: () => void;
```
Tbaut, line 14: I've a hard time figuring out how to make the best out of these, I'd remove them.

packages/extension-ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,66 @@
+// Copyright 2017-2020 @polkadot/react-components authors & contributors
```
Tbaut, line 1:
```suggestion
// Copyright 2019-2020 @polkadot/extension-ui authors & contributor
```

Tbaut:
A couple questions. I've a pretty dumb/simplistic implementation with `Errorboundary` all over.
- I'll likely refactor the component with a similar pattern `<><Header /><SomeComponnet></>`. This is the case for the `Export`, `Forget`, `Import`... The question is: Is this still the type of implementation you were thinking of?
- From what I can see, we could add some "back" button in the Fallback message to redirect to the home. Is that what you were thinking about?

jacogr:
1. We may as well warp right at the root - throw error and let it handle it there. 
2. Yes.

Tbaut:
> We may as well warp right at the root - throw error and let it handle it there. (EDIT: Here https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/Popup/index.tsx#L126-L140)

Yes, I thought we could keep the header and eventually put the cancel/back button there, but this is overly complicated.
I'll rather create a nice Error component that looks like the rest, and avoid having those `ErrorBoundary` all over.

packages/extension-ui/src/Popup/index.tsx
Tbaut: removed it.

packages/extension-ui/src/components/ErrorBoundary.tsx
Tbaut: I'll keep only what I think is useful, can bring it back if needed.

packages/extension-ui/src/partials/Header.tsx
```diff
@@ -10,7 +10,8 @@ import ArrowLeftImage from '../assets/arrowLeft.svg';
 import gear from '../assets/gear.svg';
 import plus from '../assets/plus.svg';
 import logo from '../assets/pjs.svg';
-import { Link, Svg } from '../components';
+import Link from '../components/Link';
```
Tbaut, line 13: This was needed to get the extension to build.

Tbaut:
I've been stuck for a while (and I am still). The "cancel" link to return to "/" needs to be clicked twice for users to effectively return to the home. I can't figure out why. If you can have a look maybe you'll have a hint.

packages/extension-ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,81 @@
...
+    const { error } = this.state;
+    const displayError = errorProps || error;
+
+    console.log('errorProps', errorProps);
+    console.log('error', error);
+
+    const _goHome = () => {
+      this.setState({ error: null, trigger: null });
+      window.location.hash = '/';
+    };
```
jacogr, line 60: Extract this to class-level.

packages/extension-ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,81 @@
...
+import Header from '../partials/Header';
+import ActionText from './ActionText';
+// import { ThemeProps } from '../types';
+
+interface Props extends WithTranslation {
+  children: React.ReactNode;
+  className?: string;
+  error?: Error | null;
+  onError?: () => void;
+  trigger?: unknown;
```
jacogr, line 19: We have dropped trigger

jacogr: (Can also remove it rom the state)

packages/extension-ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,81 @@
...
+    return { error };
+  }
+
+  static getDerivedStateFromProps ({ trigger }: Props, { prevTrigger }: State): State | null {
+    const newTrigger = JSON.stringify({ trigger });
+
+    return (prevTrigger !== newTrigger)
+      ? { error: null, prevTrigger: newTrigger }
+      : null;
+  }
```
jacogr, line 41: Can be removed now?

packages/extension-ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,81 @@
...
+  public componentDidCatch (): void {
+    const { onError } = this.props;
+
+    onError && onError();
+  }
+
+  public render (): React.ReactNode {
+    const { children, error: errorProps, t } = this.props;
+    const { error } = this.state;
+    const displayError = errorProps || error;
```
jacogr, line 52: This is the only place where it _may_ be an issue, but since we don't pass props through... also stumped just looking at the code.

packages/extension-ui/src/components/ErrorBoundary.tsx
Tbaut: I've been scratching my head a lot about this yesterday simply because I don't really understand it. The implementation I used before is also using some kind of comparison ([here](https://github.com/bvaughn/react-error-boundary/blob/master/src/index.js#L3-L4)) that I don't really understand.

packages/extension-ui/src/components/ErrorBoundary.tsx
Tbaut: Since the state is derived from the props, the way I understand it is that `displayError === error`. I removed it now and just use the error from the state, but I still need to click cancel twice to be redirected.

Tbaut:
The double render is not due to this PR.
There's a console.log in `ErrorBoundary` and it's printed twice (same for other components).
I'll dig a bit to see what could be responsible for that.

packages/extension-ui/src/Popup/ImportSeed.tsx
```diff
@@ -34,6 +34,8 @@ export default function Import (): React.ReactElement {
   const [name, setName] = useState<string | null>(null);
   const [password, setPassword] = useState<string | null>(null);
 
+  console.log('paf');
+
```
Tbaut, line 38:
```suggestion
```

Tbaut:
Got to the bottom of this, it must have `trigger` (just like it's done in apps [here](https://github.com/polkadot-js/apps/blob/2988915727513d08affa63dfa4a0c88ec777ff09/packages/react-components/src/ErrorBoundary.tsx#L34)), because several errors are triggered otherwise.

This is how it looks when throwing `new Error('Bla)` in the seed creation intentionally:
error-boundary

packages/extension-ui/src/Popup/index.tsx
```diff
@@ -71,6 +71,8 @@ export default function Popup (): React.ReactElement {
   const _onAction = (to?: string): void => {
     setWelcomeDone(window.localStorage.getItem('welcome_read') === 'ok');
 
+    console.log('_onAction', to);
+
```
Tbaut, line 75:
```suggestion
```

packages/extension-ui/src/components/ErrorBoundary.tsx
```diff
@@ -0,0 +1,78 @@
...
+  };
+
+  public render (): React.ReactNode {
+    const { children, t, trigger } = this.props;
+    const { error } = this.state;
+
+    if (error) {
+      console.log('trigger', trigger);
+    }
+
```
Tbaut, line 54:
```suggestion
```

jacogr:
Still click-crazy?

Tbaut:
Nope, it's good now.

jacogr (approved):
lgtm.

packages/extension-ui/src/Popup/index.tsx
```diff
@@ -103,15 +103,19 @@ export default function Popup (): React.ReactElement {
       .catch(console.error);
   }, [cameraOn]);
 
+  const wrapWithErrorBoundary = (component: React.ReactElement, trigger?: string) : React.ReactElement => {
+    return <ErrorBoundary trigger={trigger}>{component}</ErrorBoundary>;
+  };
```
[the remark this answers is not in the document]
Tbaut: Ah sorry, I usually try to follow that!

PR #473, opened 2020-09-30
Refactor open window

### LINKED ISSUE #472: Recover account from mnemonic with password manager is inpractical
Similar than #400, recovering an account with mnemonic where the seed and/or the password are stored in a password manager is annoying and must be done in a separate tab.

I'd suggest the same behaviour as JSON recovery when we automatically open a new tab.

Tbaut:
originally linked to #472 
I've refactored `windowOpen` to be generic enough to handle (securely hopefully?) any window open.


packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -377,12 +377,17 @@ export default class Extension {
     return true;
   }
 
-  private windowOpen (path = '/'): boolean {
-    console.error('open', `${chrome.extension.getURL('index.html')}#${path}`);
```
Tbaut, line 381: Any reason to have had a console.error here? I replaced it by a log.

packages/extension-ui/src/partials/MenuAdd.tsx
```diff
@@ -11,23 +11,31 @@ import plusIcon from '../assets/plus.svg';
...
 import useTranslation from '../hooks/useTranslation';
-import { jsonRestoreWindowOpen } from '../messaging';
+import { windowOpen } from '../messaging';
 
 interface Props extends ThemeProps {
   className?: string;
   reference: React.MutableRefObject<null>;
 }
 
-const isPopup = window.innerWidth <= 480;
```
Tbaut, line 22: squashed what looks like a a bug along the way, was set as 480 here and 560 below. Built a hook for this now.

jacogr: 2nd time... fixed something similar in the last round :(

packages/extension-ui/src/hooks/useIsPopup.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2019-2020 @polkadot/extension-ui authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import { useMemo } from 'react';
+
+export default function useIsPopup (): {isPopup: boolean} {
+  return useMemo(() => {
+    return { isPopup: window.innerWidth <= 560 };
+  }, []);
```
jacogr, line 9:
```suggestion
export default function useIsPopup (): boolean {
  return useMemo(() => {
    return window.innerWidth <= 560;
  }, []);
```

packages/extension-ui/src/messaging.ts
```diff
@@ -198,8 +194,8 @@ export async function deriveAccount (parentAddress: string, suri: string, parent
   return sendMessage('pri(derivation.create)', { genesisHash, name, parentAddress, parentPassword, password, suri });
 }
 
-export async function jsonRestoreWindowOpen (): Promise<boolean> {
-  return sendMessage('pri(window.open.json)', null);
+export async function windowOpen (path: AllowedPath): Promise<boolean> {
```
jacogr, line 197: Nice! It doesn't happen often that we get to remove messages.

packages/extension-ui/src/partials/MenuAdd.tsx
```diff
@@ -11,23 +11,31 @@ import plusIcon from '../assets/plus.svg';
...
   const mediaAllowed = useContext(MediaContext);
+  const { isPopup } = useIsPopup();
 
   const _openJson = useCallback((): void => {
-    jsonRestoreWindowOpen().catch(console.error);
+    windowOpen(jsonPath).catch(console.error);
+  }, []);
+
+  const _openMnemonic = useCallback((): void => {
+    console.log('goooo');
```
jacogr, line 37:
```suggestion
```

packages/extension-ui/src/partials/MenuSettings.tsx
```diff
@@ -33,6 +33,7 @@ function MenuSettings ({ className, reference }: Props): React.ReactElement<Prop
   const [prefix, setPrefix] = useState(`${settings.prefix === -1 ? 42 : settings.prefix}`);
   const themeContext = useContext<Theme>(ThemeContext);
   const setTheme = useContext(ThemeSwitchContext);
+  const { isPopup } = useIsPopup();
```
jacogr, line 36:
```suggestion
  const isPopup = useIsPopup();
```

Adjusted with hook suggestion above.

jacogr (commented):
Like the cleanups, but really not sold at all with the import opening in a new window. For JSON there is a reason since it is the only way to actually attach files.

I'm not against this functionality in the future, but as it stands the "open-in-new-window" is very limited, doesn't use the space and overall looks out of place. To fit more in there (and basically move the bulk of the operations there) detracts until fixed.

Tbaut:
Got it, I'll remove the new window there. I agree that it feels out of place. I was on the fence between this solution, and proposing yet another "cache during 2min" to redirect automatically to the seed recovery screen when exited.

The fact that we do not treat users using a password manager as first class citizens is bothering me a bit. We should not make it hard for users to have good practices. forcing them to open in a new tab is somewhat annoying.

jacogr:
Yes, the seed thing is a bit hacky. If the additions are first-class new-window everywhere that problem goes away. And it is consistent. However have def. had the feedback before in issues here that the "open window" itself is just also out of place. So that is probably the place to start.

jacogr (commented):
Looks good. A little conflict just snuck in with i18n (probably could have resolved myself, but you are better suited... just-in-case I mess up)

Tbaut:
Need to double check locally that I didn't mess up myself (I edited it on github)

Tbaut:
Looks good :)

jacogr (approved):

PR #475, opened 2020-10-01
Fix missing translations

Tbaut:
Also removes a seemingly unused component.


packages/extension/public/locales/en/translation.json
```diff
@@ -97,5 +97,6 @@
   "Understood, let me continue": "",
   "You currently don't have any accounts. Create your first account to get started.": "",
   "Please write down your wallet's mnemonic seed and keep it in a safe place. The mnemonic can be used to restore your wallet. Keep it carefully to not lose your assets.": "",
-  "You are exporting your account. Keep it safe and don't share it with anyone.": ""
+  "You are exporting your account. Keep it safe and don't share it with anyone.": "",
+  "Create an account:": ""
```
Tbaut, line 101: I don't really understand this. I tried removing it and running `yarn build:i18n` again, and it actually ended up acting even more weird, by deleting many valid entries. Could it be something wrong with the scanner config? I'll see what I can do.

jacogr (approved):
Brilliant!

packages/extension/public/locales/en/translation.json
jacogr: There should be an unused/duplicate check. I could have mucked up when I copied from apps, sadly I did it in 5 minutes batches all the way though.

Tbaut: ok I'll try to dig deeper because this is truly weird. There's no "Create an account:" in the code, there's a very similar one, without the colon at the end, and it's in the list already.

Tbaut: The problem comms from the `<string>` in `t<string>(...` because the scanner has no transform similar to https://github.com/polkadot-js/apps/blob/master/i18next-scanner.config.js#L10

I don't know exactly what this `<string>` helped for, and would be for removing it in the extension if that can make our life easier.

jacogr: Ok, I have no duplicate thing here it would seem. So what I suggest, remove the file completely, run it, then see if it appears. (That is a manual clean)

Tbaut: This is what I did, the output is the following:

```bash
i18next-scanner: options={
  "debug": true,
  "sort": false,
  "attr": {
    "list": [
      "data-i18n"
    ],
    "extensions": [
      ".html",
      ".htm"
    ]
  },
  "func": {
    "extensions": [
      ".tsx",
      ".ts"
    ],
    "list": [
      "t",
      "i18next.t",
      "i18n.t"
    ]
  },
  "trans": {
    "component": "Trans",
    "i18nKey": "i18nKey",
    "defaultsKey": "defaults",
    "extensions": [
      ".js",
      ".jsx"
    ],
    "fallbackKey": false,
    "acorn": {
      "ecmaVersion": 10,
      "sourceType": "module"
    }
  },
  "lngs": [
    "en"
  ],
  "fallbackLng": "en",
  "ns": [
    "translation"
  ],
  "defaultLng": "en",
  "defaultNs": "translation",
  "defaultValue": "",
  "resource": {
    "jsonIndent": 2,
    "lineEnding": "\n",
    "loadPath": "packages/extension/public/locales/{{lng}}/{{ns}}.json",
    "savePath": "packages/extension/public/locales/{{lng}}/{{ns}}.json"
  },
  "keySeparator": false,
  "nsSeparator": false,
  "context": true,
  "contextFallback": true,
  "contextSeparator": "_",
  "contextDefaultValues": [],
  "plural": true,
  "pluralFallback": true,
  "pluralSeparator": "_",
  "interpolation": {
    "prefix": "{{",
    "suffix": "}}"
  }
}
i18next-scanner: Added a new translation key { "Copied": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "<unknown>": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "click to select or drag and drop the file here": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "{{name}} ({{size}} bytes)": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "Create new account (root or derived)": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "Create new account": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "Derivation Path": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "Derivation Path (unlock to edit)": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "Sign the transaction": "" } to "packages/extension/public/locales/en/translation.json"
i18next-scanner: Added a new translation key { "Sign the message": "" } to "packages/extension/public/locales/en/translation.json"
```

So it adds only the strings that have no type, like https://github.com/polkadot-js/extension/blob/bc4e57e353940dbabb8a53ae7dc63e7038b924ea/packages/extension-ui/src/Popup/Signing/index.tsx#L62 
Adding the type and it disappears.

It seems sadly that TS does complain if I remove the `<string>` most of the time. So I'll add the transform like in apps.

jacogr: That is weird... it def. extracted them _with_ the type. (And yes, that stuff is new, hated adding that boilerplate everywhere, but some version of eslint with a plugin started complaining bitterly...)

Tbaut: looks sane now with 789a54e

jacogr (approved):
Cool! Perfect. Now I wish I had the "check for gaps script" here to send the delta (couple of items) for zh translation.

Tbaut:
Sure, I guess we have something similar in `apps`, else I'll write it. Doing this in another PR.

jacogr:
Not sure the apps things does the right thing here - I believe (it has been a while), it only removes "dead stuff", but _may_ alert on missing/new.

Either way, it generally is ok-ish to "keep things aligned", but doesn't help the translators at all. But it has a benefit that it has a built-in translation tool (way overkill here) that _could_ show all the new/missing stuff to look at. (It doesn't).

Either way, it is a gap. I suppose here it may just be as simple as manually comparing (it is not massive), but it really is a waste of time when you try and scale it. Alternatively need to start looking externally at some of the tools the W3F are using with JSON files.

PR #481, opened 2020-10-06
Groupe accounts by network and sort by name, path, creation date

### LINKED ISSUE #474: Better sorting, not only by created
Sorting in the extension also affects the exports, i.e. it is exported in the order displayed. However the current sorting inside the app is done on parent and then on creation only. My suggestion (open for adaptation) would be -

- top-level based on name - when same use creation timestamp
- derived accounts, based on actual name - when same, use path

Certainly would make my own life much easier.

Tbaut:
closes #474 

compare

### DIFF
--- a/packages/extension-ui/src/util/buildHierarchy.ts
+++ b/packages/extension-ui/src/util/buildHierarchy.ts
@@ -3,13 +3,15 @@
 
 import { AccountJson, AccountWithChildren } from '@polkadot/extension-base/background/types';
 
+import getNetworkMap from './getNetworkMap';
+
 type ChildFilter = (account: AccountJson) => AccountWithChildren;
 
 function compareByCreationTime (a: AccountJson, b: AccountJson): number {
   return (a.whenCreated || Infinity) - (b.whenCreated || Infinity);
 }
 
-function compareByString (a?: string, b?: string): number {
+function compareByString (a?: string | null, b?: string | null): number {
   const nameA = a?.toUpperCase() || '';
   const nameB = b?.toUpperCase() || '';
 
@@ -33,7 +35,19 @@ function compareByPath (a: AccountJson, b: AccountJson): number {
   return compareByString(a.suri, b.suri);
 }
 
+function compareByNetwork (a: AccountJson, b: AccountJson): number {
+  const networkMap = getNetworkMap();
+  const networkA = networkMap.get(a?.genesisHash || '');
+  const networkB = networkMap.get(b?.genesisHash || '');
+
+  return compareByString(networkA, networkB);
+}
+
 function compareByNameThenCreation (a: AccountJson, b: AccountJson): number {
+  if (a.genesisHash !== b.genesisHash) {
+    return 0;
+  }
+
   const res = compareByName(a, b);
 
   return res === 0 ? compareByCreationTime(a, b) : res;
@@ -67,5 +81,6 @@ export function buildHierarchy (accounts: AccountJson[]): AccountWithChildren[] 
       !accounts.some(({ address }) => parentAddress === address)
     )
     .map(accountWithChildren(accounts))
+    .sort(compareByNetwork)
     .sort(compareByNameThenCreation);
 }
--- a/packages/extension-ui/src/util/buildHierarchy.ts
+++ b/packages/extension-ui/src/util/buildHierarchy.ts
@@ -9,12 +9,28 @@ function compareByCreationTime (a: AccountJson, b: AccountJson): number {
   return (a.whenCreated || Infinity) - (b.whenCreated || Infinity);
 }
 
+function compareByNameThenTimestamp (a: AccountJson, b: AccountJson): number {
+  const nameA = a?.name?.toUpperCase() || '';
+  const nameB = b?.name?.toUpperCase() || '';
+
+  if (nameA < nameB) {
+    return -1;
+  }
+
+  if (nameA > nameB) {
+    return 1;
+  }
+
+  // names are equal, compare by timestamp
+  return compareByCreationTime(a, b);
+}
+
 export function accountWithChildren (accounts: AccountJson[]): ChildFilter {
   return (account: AccountJson): AccountWithChildren => {
     const children = accounts
       .filter(({ parentAddress }) => account.address === parentAddress)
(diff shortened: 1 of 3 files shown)


Tbaut:
Forgot the sorting by path for children, adding it now.

Tbaut:
Actually, since users can't derive freely, deriving by path or by creation date doesn't change anything since the UI obliges to go in order. I still implemented it to be future proof in case we want to let the user free to decide the derivation paths.

jacogr:
In you example above, we may want to sort by chain first before anything else.

Tbaut:
With some more tests and chain grouping now.
network

packages/extension-ui/src/util/buildHierarchy.ts
```diff
@@ -3,18 +3,70 @@
...
+function compareByNameThenCreation (a: AccountJson, b: AccountJson): number {
+  // This comparison happens after an initial sorting by network.
+  // if the 2 accounts are from different networks, don't touch their order
+  if (a.genesisHash !== b.genesisHash) {
+    return 0;
+  }
+
+  const res = compareByName(a, b);
+
+  return res === 0 ? compareByCreationTime(a, b) : res;
```
jacogr, line 55: Should move this into path since it gets done last.

packages/extension-ui/src/util/buildHierarchy.ts
```diff
@@ -3,18 +3,70 @@
...
 
 function compareByCreationTime (a: AccountJson, b: AccountJson): number {
   return (a.whenCreated || Infinity) - (b.whenCreated || Infinity);
 }
 
+function compareByString (a?: string | null, b?: string | null): number {
+  const nameA = a?.toUpperCase() || '';
+  const nameB = b?.toUpperCase() || '';
+
+  if (nameA < nameB) {
```
jacogr, line 18: `nameA.localeCompare(nameB)`?

packages/extension-ui/src/util/buildHierarchy.ts
```diff
@@ -3,18 +3,70 @@
...
+  if (nameA > nameB) {
+    return 1;
+  }
+
+  // names are equal
+  return 0;
+}
+
+function compareByName (a: AccountJson, b: AccountJson): number {
+  return compareByString(a.name, b.name);
```
jacogr, line 31: Can actually just move the localCompare in here.

packages/extension-ui/src/util/buildHierarchy.ts
Tbaut: thanks, didn't know about this one.

packages/extension-ui/src/util/buildHierarchy.ts
Tbaut: Yes, if we want to apply the same type of comparison Name > Path > Creation for each. I had Name > Creation for root and Name > Path for children (no Creation). Will change that then.

packages/extension-ui/src/util/buildHierarchy.ts
```diff
@@ -3,18 +3,60 @@
...
+function compareByNameThenPathThenCreation (a: AccountJson, b: AccountJson): number {
+  // This comparison happens after an initial sorting by network.
+  // if the 2 accounts are from different networks, don't touch their order
+  if (a.genesisHash !== b.genesisHash) {
+    return 0;
+  }
+
+  const res = compareByName(a, b);
+
+  return res === 0 ? compareByPathThenCreation(a, b) : res;
```
jacogr, line 51:
```suggestion
  return compareByName(a, b) || compareByPathThenCreation(a, b);
```

packages/extension-ui/src/util/buildHierarchy.ts
```diff
@@ -3,18 +3,60 @@
...
+  const networkA = networkMap.get(a?.genesisHash || '') || '';
+  const networkB = networkMap.get(b?.genesisHash || '') || '';
+
+  return networkA.localeCompare(networkB);
+}
+
+function compareByPathThenCreation (a: AccountJson, b: AccountJson): number {
+  const res = compareByPath(a, b);
+
+  return res === 0 ? compareByCreation(a, b) : res;
```
jacogr, line 39:
```suggestion
  return compareByPath(a, b) || compareByCreation(a, b);
```

jacogr (approved):
lgtm, 2 small comments that may help or may hinder.

packages/extension-ui/src/util/buildHierarchy.ts
Tbaut: I'll add a comment just to make it clear since I find this quite hard to understand at first glance.

packages/extension-ui/src/util/buildHierarchy.ts
```diff
@@ -3,18 +3,56 @@
...
+}
+
+function compareByNameThenPathThenCreation (a: AccountJson, b: AccountJson): number {
+  // This comparison happens after an initial sorting by network.
+  // if the 2 accounts are from different networks, don't touch their order
+  if (a.genesisHash !== b.genesisHash) {
+    return 0;
+  }
+
+  return compareByName(a, b) || compareByPathThenCreation(a, b);
```
Tbaut, line 47:
```suggestion
  // if the names are equal, compare by path then creation time
  return compareByName(a, b) || compareByPathThenCreation(a, b);
```

packages/extension-ui/src/util/buildHierarchy.ts
```diff
@@ -3,18 +3,56 @@
...
+function compareByNetwork (a: AccountJson, b: AccountJson): number {
+  const networkMap = getNetworkMap();
+  const networkA = networkMap.get(a?.genesisHash || '') || '';
+  const networkB = networkMap.get(b?.genesisHash || '') || '';
+
+  return networkA.localeCompare(networkB);
+}
+
+function compareByPathThenCreation (a: AccountJson, b: AccountJson): number {
+  return compareByPath(a, b) || compareByCreation(a, b);
```
Tbaut, line 37:
```suggestion
  // if the paths are equal, compare by creation time
  return compareByPath(a, b) || compareByCreation(a, b);
```

PR #483, opened 2020-10-06
Fix derive suggestion when switching root

### LINKED ISSUE #482: Deriving an account using the "Create account" shows a wrong derivation path suggestion
I have an account A, with a child with suri `/0`.
Using the "Create new account", and changing to this account, the suggested path is `/0` although I already have such a child account.
Using the dedicated menu for this account shows `/1` as suggestion correctly

Tbaut:
closes #482

There doesn't seem to be a way to test this correctly since this involves a redirection.
derive-2


jacogr (approved):
Thank you!

PR #484, opened 2020-10-07
Show a specific icon on external account

### LINKED ISSUE #480: See at glance if an account is external
I'd like to be able to see quickly if accounts are stored externally from the extension (signer or ledger) or if the extension has the keys. I imagine that it could materialize as an icon next to the identicon or the name.

Tbaut:
closes #480 

How it looks like:
external

### DIFF
--- a/packages/extension-ui/src/Popup/Forget.tsx
+++ b/packages/extension-ui/src/Popup/Forget.tsx
@@ -1,6 +1,5 @@
 // Copyright 2019-2020 @polkadot/extension-ui authors & contributors
 // SPDX-License-Identifier: Apache-2.0
-import { AccountJson } from '@polkadot/extension-base/background/types';
 
 import React, { useCallback, useContext, useMemo } from 'react';
 import { RouteComponentProps, withRouter } from 'react-router';
--- a/packages/extension-ui/src/Popup/Forget.tsx
+++ b/packages/extension-ui/src/Popup/Forget.tsx
@@ -1,11 +1,11 @@
 // Copyright 2019-2020 @polkadot/extension-ui authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import React, { useCallback, useContext, useMemo } from 'react';
+import React, { useCallback, useContext } from 'react';
 import { RouteComponentProps, withRouter } from 'react-router';
 import styled from 'styled-components';
 
-import { AccountContext, ActionContext, Address, ActionBar, ActionText, Button, Warning } from '../components';
+import { ActionContext, Address, ActionBar, ActionText, Button, Warning } from '../components';
 import useTranslation from '../hooks/useTranslation';
 import { forgetAccount } from '../messaging';
 import { Header } from '../partials';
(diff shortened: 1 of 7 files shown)


packages/extension-ui/src/Popup/Forget.tsx
```diff
@@ -15,6 +15,22 @@ type Props = RouteComponentProps<{ address: string }>;
...
+        res = account.isExternal;
+
+        return true;
+      }
+
+      return false;
+    });
+
+    return res;
+  }, [accounts, address]);
```
jacogr, line 33:
```suggestion
  const isExternal = useMemo(
    () => accounts.some((account) =>
      account.address === address && account.isExternal
    )),
    [accounts, address]);
```

I understand the above, mapping to address first, but it seems weird to me. If that approach is really to be followed, either filter + isExternal or find + isExternal.

jacogr (commented):
Not sure about the placement? Possibly next to the invisible icon?

Tbaut:
I gave it a shot next to the copy button and the visible indicator.
Placing it here would mean significantly increasing the card size (from 72px to 90+) also the copy and visibility icons are actionable (well the visibility isn't yet, but I think it should, I'll log it) whereas the fact that an account is external is definitely not.
 
image

I took this idea form [fether](https://user-images.githubusercontent.com/6226175/50950496-40079c00-14fe-11e9-9c14-6cb496bf658f.png)

packages/extension-ui/src/Popup/Forget.tsx
Tbaut: thanks, I used find.

jacogr:
Next to the visibility icon? Same line? Agreed that we don't want to make the accounts higher. 

As you suggested elsewhere, we may want to keep that always visible (strike-through vs non) so the external doesn't move around.

Tbaut:
Looks weird to me
image


I'll check it out next to the name

Tbaut:
Here with the name before
image

after
image

Tbaut:
I think my favourite is the arrow before the name. What's your opinion?

jacogr:
Yes, think it works. Would just adjust it to be the same lightness as the visibility one.

packages/extension-ui/src/Popup/Forget.tsx
```diff
@@ -36,7 +36,9 @@ function Forget ({ match: { params: { address } } }: Props): React.ReactElement<
         text={t<string>('Forget account')}
       />
       <div>
-        <Address address={address}>
+        <Address
+          address={address}
+        >
```
Tbaut, line 41:
```suggestion
        <Address address={address}>
```

packages/extension-ui/src/components/Address.tsx
```diff
@@ -237,6 +251,7 @@ const Settings = styled.div(({ theme }: ThemeProps) => `
     cursor: pointer;
     background: ${theme.readonlyInputBackground};
   }
+
```
Tbaut, line 254:
```suggestion
```

Tbaut:
- Set the color as the label (same as the visibility ones when hovered)
- Added a title attribute, it will surely help some

Here's how it looks now:
external

Selection_266

jacogr:
Ok, let's get it in. I would suggest (once we have ledger) to just make the icons different, QR for the one, something hardware-related for the next.

jacogr (approved):
lgtm

Tbaut:
> Ok, let's get it in. I would suggest (once we have ledger) to just make the icons different, QR for the one, something hardware-related for the next.

Sure

PR #488, opened 2020-10-12
Add phishing site detection, redirection and information

### LINKED ISSUE #464: Check for Polkadot phishing sites
... redirect when detected (Metamask does this really well)

Tbaut:
closes #464 

image

Comments / questions
- not happy with the `View` that makes the text shrinked on a 560px centered window. The most straight forward solution I thought about (not great though) was to use a bare html page to render without React, but that's additional styling etc to maintain, so left it as is for now.
- Unlike Metamask, I did not want to add a button to `still allow this website`.

### DIFF
--- a/packages/extension-base/src/background/types.ts
+++ b/packages/extension-base/src/background/types.ts
@@ -108,6 +108,7 @@ export interface RequestSignatures {
   'pub(extrinsic.sign)': [SignerPayloadJSON, ResponseSigning];
   'pub(metadata.list)': [null, InjectedMetadataKnown[]];
   'pub(metadata.provide)': [MetadataDef, boolean];
+  'pub(phishing.redirect)': [null, null];
   'pub(rpc.listProviders)': [void, ResponseRpcListProviders];
   'pub(rpc.send)': [RequestRpcSend, JsonRpcResponse];
   'pub(rpc.startProvider)': [string, ProviderMeta];
@@ -351,3 +352,7 @@ export interface ResponseJsonRestore {
 }
 
 export type AllowedPath = typeof ALLOWED_PATH[number];
+
+export interface RequestPhishingRedirect {
+  host: string;
+}
--- a/packages/extension-base/src/utils/isXmlOrPdf.ts
+++ b/packages/extension-base/src/utils/isXmlOrPdf.ts
@@ -4,7 +4,8 @@
 export default function (pathname: string): boolean {
   const prohibitedFiles: RegExp[] = [
     /\.xml$/u,
-    /\.pdf$/u
+    /\.pdf$/u,
+    /1281/u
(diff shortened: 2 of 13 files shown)


packages/extension-base/src/background/types.ts
```diff
@@ -351,3 +352,7 @@ export interface ResponseJsonRestore {
 }
 
 export type AllowedPath = typeof ALLOWED_PATH[number];
+
+export interface RequestPhishingRedirect {
+  host: string;
+}
```
Tbaut, line 358: stray interface.
```suggestion
```

jacogr:
Ok, the PDF/XML check seems useful. I believe the extension may want to run on those pages? Although the should not try and run? Not sure there is an issue (that I've seen) to be fixed here, but can be wrong.

packages/extension-base/src/utils/isXmlOrPdf.ts
```diff
@@ -0,0 +1,19 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+export default function (pathname: string): boolean {
+  const prohibitedFiles: RegExp[] = [
+    /\.xml$/u,
+    /\.pdf$/u
+  ];
+
+  return prohibitedFiles.some((file) => {
+    console.log(pathname, file, file.test(pathname));
```
Tbaut, line 11: stray log
```suggestion
```

packages/extension-ui/src/Popup/PhishingDetected.tsx
```diff
@@ -16,30 +16,28 @@ export default function PhishingDetected (): React.ReactElement<Props> {
...
-          Just because a domain does not appear on this list is not an implicit guarantee of that domain's safety.
-          As always, your transactions are your own responsibility.
+          This redirection is based on a list of websites accessible at https://github.com/polkadot-js/phishing. Note that this is a community-driven, curated list.
+          It might be incomplete or inaccurate.
       `)}
       </Note>
       <Note>
         {t<string>(`
-          If you think this domain is incorrectly flagged or if a blocked legitimate website has resolved its security issues,
+          If you think that this website was flagged incorrectly,
```
jacogr, line 35: For these, use the `<Trans />` component from `i18next`. From memory.

```js
<Trans key='phishing.incorrect'>If you think that this website was flagged incorrectly,
  <IssueLink href='https://github.com/polkadot-js/phishing/issues/new'>please open an issue by clicking here</IssueLink></Trans>
```

The reason is not all languages follow the same structure

jacogr:
The look is fine for now. Not great, but it has the same issue as elsewhere - we actually need a full-screen view at some point.

jacogr (approved):
lgtm.

PR #489, opened 2020-10-13
Password caching refactoring

Tbaut:
- renaming to make things clear `savePass` is an action, whether the password is needed or not is told by the `isLocked` on the pair
- use an "updateCache" function that is called
  - when the Request popup loads (it is part of the flow when the UI calls `isLocked`)
  - right before signing to avoid the security issue of having the account unlocked potentially longer than the set period.
  - edit the Popup is aware of the cache duration and starts a timeout until the account gets locked. 
- ~~I've been testing the event when you open the popup request before the cache expires (no password needed) and wait until it expires --> signing will fail. The experience would be much better if we could update the UI. This is non trivial though because it would mean changing this `isLocked` from a pull to a subscription. Happy to do it in this PR, it's still small.~~
edit: addressed in https://github.com/polkadot-js/extension/pull/489/commits/72564b3314a213c456c1e03b2fa9ea8d9fd8ceb7

### DIFF
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -318,7 +318,9 @@ export default class Extension {
     }
 
     // if the keyring pair is locked, the password is needed
-    assert(!pair.isLocked || !!password, 'Password needed to unlock the account');
+    if (pair.isLocked && !password) {
+      reject(new Error('Password needed to unlock the account'));
+    }
 
     if (pair.isLocked) {
       pair.decodePkcs8(password);
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -3,7 +3,7 @@
 
 import { MetadataDef } from '@polkadot/extension-inject/types';
 import { SubjectInfo } from '@polkadot/ui-keyring/observable/types';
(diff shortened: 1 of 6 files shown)


packages/extension-base/src/defaults.ts
```diff
@@ -5,7 +5,7 @@ const ALLOWED_PATH = ['/', '/account/restore-json'] as const;
 const PORT_CONTENT = 'content';
 const PHISHING_PAGE_REDIRECT = '/phishing-page-detected';
 const PORT_EXTENSION = 'extension';
-const PASSWORD_EXPIRY_MIN = 15;
+const PASSWORD_EXPIRY_MIN = 1;
```
Tbaut, line 8: for testing, needs to be set back.

jacogr:
Ahh, I see what you mean. So while waiting for a reset it fails.

So probably 2 things -

- we probably want a "reset to locked" timer in the background page (while we still have one :)
- maybe extend the error being returned to indicate it was not unlocked anymore? Do return the info on request, still a bit jarring, but it is almost an "edge-case"... dunno.

Tbaut:
> we probably want a "reset to locked" timer in the background page (while we still have one :)

Do you mean to add a timer in place of something else? Because unless there's a subscription to the locking event, the background timer will not help much. Also as you mention, it adds tech dept that will need to be catered for soonish with the background refactoring.

> maybe extend the error being returned to indicate it was not unlocked anymore?

Sure, I'd love to prevent any error although it's definitely an edge case.

Another idea that I had, that could have the best of both worlds:
When the request Popup loads, ask the background if unlocked (this we do already) but also return  the remaining time until it'll be locked. This allows to start a timer in the Popup, to show the password request. This would allows to:
- keep the current logic (no background timer, locking happens when calling refresh)
- prevent any signing error, hence a much better UX because the Popup has the locking time info

What do you think?

jacogr:
That works.

Tbaut:
This is def. much better in terms of user experience. Below a demo (timoute takes ~20 sec to kick in). I enter a wrong password first (a better error would be good here, I'll log that), then the right password.

signing

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -287,33 +303,41 @@ export default class Extension {
...
 
       return false;
     }
 
-    const now = Date.now();
+    this.refreshAccountPasswordCache(pair);
+
+    // if the keyring pair is locked, the password is needed
+    if (pair.isLocked && !password) {
+      reject(new Error('Password needed to unlock the account'));
```
Tbaut, line 329: BTW, unlike what I said, I actually can't give a custom error here because the rejection does not only occur when the cache expires, it could also simply be an empty password (although this shouldn't be possible any more since I disabled the button).

jacogr (approved):
lgtm.

Issue #494, opened 2020-10-14
Show an explicit error message upon signing reuqest with wrong password


Tbaut:
Right now the only pointer is that the password field is reset and the ticker is red.
This is a wrong password now:
![image](https://user-images.githubusercontent.com/33178835/95982509-7afbd980-0e20-11eb-8997-fa32b77a16bd.png)

tomusdrw:
Is this already deployed? I recently run into this and only had an indication in the console (extension downloaded today from Chrome WebStore).

Tbaut:
You guessed it, it's part of v0.35 that isn't out yet.

Issue #491, opened 2020-10-13
Checking the "remember for X min" checkbox resets the password


Tbaut:
How to reproduce:
- See the request pop up
- type in your password
- click the checkbox (it's after the password input)
--> password is reset for no good reason

Issue #487, opened 2020-10-09
Remeber Me -Polkadot Extension timing 


arunrajansharma:
I would like to request the team to implement a prolonged 15 min remember e.g if I have made a transaction in a 15 min time window, then the extension will not ask me for the next 15 mins. The current setup is quite cumbersome, requiring me to put my password every 15 mins. Thanks

siman:
+1. This is very needed for Subsocial

jacogr:
Actually had it this way around in the initial implementation, but ran out of time since it needed to go due to API updates - so the idea was to keep the checkbox there (and checked) when you sign. So each time you prolong. when unchecked, it forgets.

(Initial is still opt-in). 

Overall like it and it doesn't detract from security, nor add any additional (mostly unneeded noise).

siman:
Amazing news :) thank you for good feedback, @jacogr

siman:
@Tbaut any other ideas how to improve UX of extension and do not see a tx confirmation modal every single time? People keep complaining about this confirmation modal. Nobody wants to see it so often. Me included

Tbaut:
Just to be clear, this has nothing to do with the UX of this extension. The "People keep complaining" certainly about the UX of a web3 application, which is different, and I'm happy to discuss since I'm also impacted as a user. This issue is not the right place though.

Making the web3 account keeping insecure is certainly not the best way to reach a broader audience and make the Dapps' UX better. 

My [comment](https://github.com/polkadot-js/extension/issues/487#issuecomment-707756110) is a heads up. Prolonging the 15min has been accepted above by Jaco, and I will submit a PR soon. This will not get rid of the Popup since we still want user approval.

subsocialdev:
> Making the web3 account keeping insecure is certainly not the best way to reach a broader audience and make the Dapps' UX better.

@Tbaut what good UX solutions to this problem have you seen?

PR #503, opened 2020-10-15
Refactor password error in request pop up and remember checkbox

Tbaut:
closes https://github.com/polkadot-js/extension/issues/487
closes https://github.com/polkadot-js/extension/issues/494
closes https://github.com/polkadot-js/extension/issues/491

- checkbox checked by default to extend the "password remember" period. Wording suggestion welcome :)
- unchecking this checkbox will lock the account after this last signature (so it's either extend, or lock after this one)

Here is a demo
- I first enter a wrong password twice, then type the right one and set to remember the password for 15min
- submit another tx, the password isn't needed, the checkbox to extend is checked, I uncheck it
- submit another tx -> password is asked.

remember


jacogr (approved):
Thank you!

Issue #496, opened 2020-10-14
Replace usage of extension-chains with @polkadot/networks


jacogr:
https://github.com/polkadot-js/common/pull/707

jacogr:
Example in the apps UI - https://github.com/polkadot-js/apps/pull/3876

jacogr:
Probably not "replace", rather enhance. Basically this needs to be replaced by a common source - https://github.com/polkadot-js/extension/blob/master/packages/extension-chains/src/chains.ts

jacogr:
Also be aware of https://github.com/polkadot-js/common/pull/710 (array of genesisHash, needed in the keyring replacement)

PR #504, opened 2020-10-16
Use @polkadot/networks

Tbaut:
closes https://github.com/polkadot-js/extension/issues/496

### DIFF
--- a/packages/extension-chains/src/chains.ts
+++ b/packages/extension-chains/src/chains.ts
@@ -2,64 +2,15 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import { MetadataDefBase } from '@polkadot/extension-inject/types';
+import { all } from '@polkadot/networks';
 
-const hashes: MetadataDefBase[] = [
-  // keep at the top (favoritism)
-  {
-    chain: 'Polkadot',
-    genesisHash: '0x91b171bb158e2d3848fa23a9f1c25182fb8e20313b2c1eb49219da7a70ce90c3',
-    icon: 'polkadot',
-    ss58Format: 0
-  },
-  {
-    chain: 'Kusama',
-    genesisHash: '0xb0a8d493285c2df73290dfb7e61f870f17b41801197a149ca93654499ea3dafe',
-    icon: 'polkadot',
-    ss58Format: 2
-  },
-  // alphabetical to avoid (further) favorites
-  {
-    chain: 'Crab',
-    genesisHash: '0x34f61bfda344b3fad3c3e38832a91448b3c613b199eb23e5110a635d71c13c65',
-    icon: 'substrate',
-    ss58Format: 42
-  },
-  {
-    chain: 'Darwinia',
-    genesisHash: '0x729cb8f2cf428adcf81fe69610edda32c5711b2ff17de747e8604a3587021db8',
-    icon: 'substrate',
-    ss58Format: 18
-  },
-  {
-    chain: 'Edgeware',
-    genesisHash: '0x742a2ca70c2fda6cee4f8df98d64c4c670a052d9568058982dad9d5a7a135c5b',
-    icon: 'substrate',
-    ss58Format: 7
-  },
-  {
-    chain: 'Kulupu',
-    genesisHash: '0xf7a99d3cb92853d00d5275c971c132c074636256583fee53b3bbe60d7b8769ba',
-    icon: 'substrate',
-    ss58Format: 16
-  },
-  {
-    chain: 'Plasm',
-    genesisHash: '0x3e86364d4b4894021cb2a0390bcf2feb5517d5292f2de2bb9404227e908b0b8b',
-    icon: 'substrate',
-    ss58Format: 5
-  },
-  {
-    chain: 'Stafi',
-    genesisHash: '0x290a4149f09ea0e402c74c1c7e96ae4239588577fe78932f94f5404c68243d80',
-    icon: 'substrate',
-    ss58Format: 20
-  },
-  {
-    chain: 'Subsocial',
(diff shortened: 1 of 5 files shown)


packages/extension-chains/src/chains.ts
```diff
@@ -2,64 +2,15 @@
...
-  },
-  {
-    chain: 'Subsocial',
-    genesisHash: '0x0bd72c1c305172e1275278aaeb3f161e02eccb7a819e63f62d47bd53a28189f8',
-    icon: 'substrate',
-    ss58Format: 28
-  }
-];
+const hashes: MetadataDefBase[] =
+all.filter(({ genesisHash }) => !!genesisHash)
```
jacogr, line 8: Just use `known` and don't apply filters. (That aligns with what we have in apps, so the list between the 2 always aligns). Additionally don't need to remove "relay chain", let's just keep it to the list. 

(If that impacts the network display label top-right, that will need to be filtered there)

Tbaut: Thanks I missed `known`.
Yes the "Relay Chain" make labels too long, I'll replace later and modify the dropdown to forget about (live / canary) because it's overflowing also.

packages/extension-chains/src/chains.ts
```diff
@@ -2,15 +2,14 @@
...
-all.filter(({ genesisHash }) => !!genesisHash)
-  .map((network) => ({
-    chain: network.displayName.replace('Relay Chain', ''),
-    genesisHash: (network.genesisHash as string[])[0],
-    icon: network.icon || 'jdenticon',
-    ss58Format: network.prefix
-  }));
+networks.map((network) => ({
+  chain: network.displayName,
+  genesisHash: network.genesisHash?.[0] || '0x',
```
jacogr, line 10: With the current version (.17 I believe) genesisHash is always defined. (However, it may be an empty array, so `network.genesisHash[0] || '0x')

packages/extension-chains/src/chains.ts
```diff
@@ -2,15 +2,14 @@
...
+  icon: network.icon || 'jdenticon',
```
jacogr, line 11: It will always be set. (as per latest)

packages/extension-ui/src/components/Address.tsx
```diff
@@ -171,7 +171,7 @@ function Address ({ actions, address, children, className, genesisHash, isExtern
                   : undefined
               }
             >
-              {chain.name}
+              {chain.name.replace('Relay Chain', '')}
```
jacogr, line 174: +1

packages/extension-chains/src/chains.ts
Tbaut: I've the following type with the latest (this one)
```
export interface Network extends NetworkFromSubstrate {
    genesisHash: string[] | null;
    icon: Icon;
}
```

packages/extension-chains/src/chains.ts
```diff
@@ -2,15 +2,14 @@
...
 
 const hashes: MetadataDefBase[] =
-all.filter(({ genesisHash }) => !!genesisHash)
-  .map((network) => ({
-    chain: network.displayName.replace('Relay Chain', ''),
-    genesisHash: (network.genesisHash as string[])[0],
-    icon: network.icon || 'jdenticon',
-    ss58Format: network.prefix
-  }));
+networks.map((network) => ({
```
jacogr, line 8: Would just move this up to next to the equals on the previous line.

packages/extension-chains/src/chains.ts
Tbaut: yup thanks.

packages/extension-chains/src/chains.ts
jacogr: https://github.com/polkadot-js/common/blob/master/packages/networks/src/types.ts#L18

jacogr: The only one in that list that won't have a genesisHash is 42, which is the default prefix (but not really a chain)

Tbaut: ah didn't think we'd use beta for this package. Pulling that one :)

packages/extension-chains/src/chains.ts
```diff
@@ -2,64 +2,13 @@
...
-    ss58Format: 20
-  },
-  {
-    chain: 'Subsocial',
-    genesisHash: '0x0bd72c1c305172e1275278aaeb3f161e02eccb7a819e63f62d47bd53a28189f8',
-    icon: 'substrate',
-    ss58Format: 28
-  }
-];
+const hashes: MetadataDefBase[] = networks.map((network) => ({
```
jacogr, line 7: Looking at the original, instead of setting the hash to '0x` (which is the 42), we should just filter it -

- filter by genesisHash.length
- map

packages/extension-chains/src/chains.ts
Tbaut: > The only one in that list that won't have a genesisHash is 42, which is the default prefix (but not really a chain

yeah I figured out, I've put "0x" here to keep the ordering because en empty string would have put the generic accounts at the bottom, which is not what we want.

packages/extension-chains/src/chains.ts
Tbaut: I'm not sure I follow you there. Do you mean filtering it out? We actually need the 42 because this will be used by the accounts without network. Or did I miss something?

Tbaut: ok we add it elsewhere it seems, I though it was in there originally. Filtering it then :)

jacogr (approved):
Thank you!

PR #506, opened 2020-10-16
Fix scanner config (parse .ts files)

jacogr:
Same fix as in apps.

This config probably needs to move into dev and can then just be included (like we do for eg babel/jest/ts/etc)


jacogr:
This makes everything break.

PR #507, opened 2020-10-18
Some cleanups to web3Enable (god-function)

jacogr:
Better... not great.


packages/extension-dapp/src/util.ts
```diff
@@ -3,12 +3,10 @@
 
 export function documentReadyPromise <T> (creator: () => Promise<T>): Promise<T> {
   return new Promise((resolve): void => {
-    if (['complete', 'interactive'].includes(document.readyState)) {
+    if (document.readyState === 'complete') {
```
jacogr, line 6: This has been giving me endless issues in apps - something ready, sometimes not. With only complete it seems consistent.

Tbaut: Thanks for the comment, it makes this PR digestible :)

jacogr: Still having some issues atm... it is _much_ better, but it still missing initializations on some refreshes.

PR #508, opened 2020-10-18
Bump deps

jacogr:
Additionally move deps to the correct places


Tbaut:
Noted for the deps in the wrong package, thanks.

jacogr:
It shouldn't make a difference, but yarn does start complaining (rightly so) when not where they are used. But generally a nightmare, since a `yarn add` adds them to the root, so it is either more flags or a manual move after. I can't remember flags, so I move manually...

PR #511, opened 2020-10-20
Remove scrollbar on signing popup

Tbaut:
Simple, yet triggering my OCD each time :)
Tested that the scrollbar appears with longer content on FF and Chrome.

This make it look like the screenshot in #510
image
With longer content (manually added many lines to simulate a longer content)
image

instead of having disabled scrollbar like before:
image

### DIFF
--- a/packages/extension-ui/src/components/Table.tsx
+++ b/packages/extension-ui/src/components/Table.tsx
@@ -31,7 +31,7 @@ export default React.memo(styled(Table)(({ theme }: ThemeProps) => `
 
   &.isFull {
     height: 100%;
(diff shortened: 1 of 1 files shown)


jacogr:
Why the multiple lifetimes in the shot above? That looks like a bug somewhere.

Tbaut:
ah no that's a hack from me, should have mentioned that :)

jacogr (approved):
The before looks ever worse, vertical & horizontal (the horizontal should never be there)

Tbaut:
yeah that horizontal was really ugly :smile:

PR #512, opened 2020-10-20
A  simple landing page

### LINKED ISSUE #497: Add index.html page for site, update polkadot.js.org/index.html links
From https://polkadot.js.org/ we link directly here, it would be good to have a simple one-page landing site since it is user-facing

Tbaut:
closes #497 

Not sure where you want to let it live. I know that github pages usually let you choose `root` or `/docs`
Local path to the fonts didn't work locally, I expect it to work with a proper server though :man_shrugging: 

How it looks like:
landing


jacogr:
Docs is perfect. I‘ll update the config.

PR #516, opened 2020-10-21
Remove deprecated util-crypto setSS58Format

Tbaut:
Not 100% aware of the implication of this, but from what I could see there are no other implication.
We are passing the account meta (including the relevant `genesisHash`) in all our account creation methods (checked mnemonic, derive and JSON). Otherwise it'll default to substrate (42). All in all, it seems that it is only a UI thing.


jacogr:
Should be fine - just double-check the global setting with this, i.e.

- add an account not attached to a network
- change the prefix under settings, e.g. Polkadot
- that address should now be prefixed with 1 instead of 1

Think it should be fine, this is ancient before we started tying accounts to networks, so as suggested, I believe we now explicitly format each address as specified

jacogr (approved):

Tbaut:
Thanks, yup that's the kind of test I did.

PR #518, opened 2020-10-21
Style component global cleanup

Tbaut:
- Not touching https://github.com/polkadot-js/extension/blob/bc4e57e353940dbabb8a53ae7dc63e7038b924ea/packages/extension-ui/src/components/TextInputs.ts since I would need to duplicate the styling, I don't think it would make sense.

PR #520, opened 2020-10-23
Set popup window height to 600px

Tbaut:
I remember there was something special about this window height, but I can recall what it was and the internet didn't give me the answer.

In any case, this fix removes this white band at the bottom of any popup. It hadn't attract my eye so far.. and I thought I introduced it in #518 but it's also there on the live 0.34..

image


jacogr:
So the window height issue is just the max that is limited. I beleve 621 was around the limit. (There was 1px added for FF.)

jacogr (approved):

PR #522, opened 2020-10-23
Fix error field being shown before entering any name

### LINKED ISSUE #517: Account creation: Error message displayed although nothing was types yet
I thought I saw a test for this, but I might be wrong.
image

Tbaut:
closes #517 

Took me forever to notice the missing `update()` in tests and make it fail.

Interestingly, the `return` added was part of the original PR https://github.com/polkadot-js/extension/pull/299/files#diff-5ba6606e2d09b06b1994e7f7388db063bf6bb3c10bc603dd5b097bcbb61da6f9R31 and github blame doesn't show me when this was deleted, it'd be a valuable information to know if it's a mistake or if there was a good reason to do so.

### DIFF
--- a/packages/extension-ui/src/Popup/CreateAccount/CreateAccount.test.tsx
+++ b/packages/extension-ui/src/Popup/CreateAccount/CreateAccount.test.tsx
@@ -85,6 +85,7 @@ describe('Create Account', () => {
       check(wrapper.find('input[type="checkbox"]'));
       wrapper.find('button').simulate('click');
       await act(flushAllPromises);
+      wrapper.update();
     });
 
     it('only account name input is visible at first', () => {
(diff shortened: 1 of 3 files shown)


jacogr:
This whole wasMounted is weird. Will need to digest this and see real intent here, but my kneeler is that something like the mountedRef hook in apps is actually the right way to do „is mounted stuff“. 

But will take a look later today to just understand, may think it is more complicated here than it actually is.

Tbaut:
Indeed, such a tracker should actually be a `useRef` to avoid a useless re-rendering when changing the state.

jacogr:
So overall, think it fits the bill, but let's add a more correct way in here, with a mountedRef, sopied from elsewhere -

```
import { useEffect, useRef } from 'react';

export type MountedRef = React.MutableRefObject<boolean>;

export default function useIsMountedRef (): MountedRef {
  const isMounted = useRef(false);

  useEffect((): () => void => {
    isMounted.current = true;

    return (): void => {
      isMounted.current = false;
    };
  }, []);

  return isMounted;
}
```

This means we can do a ref check before making the call as well.

Tbaut:
Thanks for the code, I went for something that I find easier to read and less error prone, by returning a boolean.

jacogr (approved):
Thank you!

PR #523, opened 2020-10-24
Ability to set account visibility from main screen

### LINKED ISSUE #485: The visibility icon seems actionnable (like the copy button) but it isn't
2 possibilities:
- keep the current way of setting an account visible/invisible in the menu, the icon would only appear when not visible, clicking on it would make the account visible again.

- remove the visibility toggle from the menu setting, and show the eye icon on all accounts. Clicking on it would toggle the visibility. The obvious drawback of this solution is that it'll be a little heavier visually since we add another indicator.

image

What do you think @jacogr

Tbaut:
closes #485 

visibility


jacogr (approved):
Looks good.

PR #527, opened 2020-10-28
Show error on wrong parent password when deriving

Tbaut:
- Fix a bug where the button was blocked in the busy state indefinitely when entering a wrong password
- Show an error when the account password is wrong (was only colored red without label)
- Show an error when the derivation path is wrong
- Add/fix tests for the above scenario

derivation-password


Tbaut:
Added a description to this issue.
@jacogr this is ready to review when you have a gap.

jacogr:
This is the last thing outstanding for a release.

My only concern atm is that the "make the window smaller" thing has had a negative impact on my signing atm, have not had a chance to look. (The smaller window size there now makes the Reject look untity, half-way out of the window)

Tbaut:
Ah, didn't notice that, I'll check it out as well.

jacogr (approved):
Thank you!

PR #529, opened 2020-10-30
Add french translation

Tbaut:
Looks bad on top of master unfortunately because the font doesn't support accents. Fixing this in https://github.com/polkadot-js/extension/pull/530.

image


jacogr:
Can you re-look now that the update is in?

Would that affect apps where the same font is being PR-ed? (All 400's there)

jacogr:
I believe the do ta added are only limited weights. Salt each weight is a font pack so it is expensive. 

So it basically will round to the nearest. 

(On apps trying for on weight only - smaller, less moving parts, more consistency).

jacogr (approved):
Thank you.

PR #531, opened 2020-11-02
Use extension in iframe

zzcwoshizz:
The extension is not working in iframe. Add ```all_frames:true``` to resolve this.

### DIFF
--- a/packages/extension/manifest.json
+++ b/packages/extension/manifest.json
@@ -17,7 +17,8 @@
   "content_scripts": [{
     "js": ["content.js"],
     "matches": ["http://*/*", "https://*/*"],
-    "run_at": "document_start"
+    "run_at": "document_start",
+    "all_frames": true
   }],
   "icons": {
     "16": "images/icon-16.png",


jacogr:
Thank you for this. Don't have any strong feelings on this one way or another, however I am generally a bit worried to extend the manifest requirements since they all come with extra hoops on at a Chrome store approval level. 

However, will take a look at that, specifically on the actual "tell us why" requirements on that store and do a bit of light reading as well.

TL;DR While seemingly small, need to do some homework on my side to ensure the process is still smooth.

jbargu:
@jacogr Any progress on this?

jacogr:
No progress. Same concerns as to the absolute nightmare this opens trying to get it into stores with extra hoops and extra requirements.

So don’t expect this to be merged anytime soon.

jacogr:
I'm not going to merge this one, specifically around my concerns about the reviews each an every additional permission gets scruitinized on each new release. (As it stands this manual overhead is not low)

With that said, quite happy to re-visit when the V3 manifest goes in. I think for that (where the matching is done differently), it would be ok and generally promises to not have the same store review pains as with V2.

PR #541, opened 2020-11-23
Window height back to 621 and set 100%

Tbaut:
following #520 

- The external window is now back to 621px, the popup is still 600px as before.
- I've set a couple `height:100%` to prevent the empty space at the bottom that made me propose #520. 
- This `min-height 600px` was necessary for the poup to be at 600 (it was all shrink otherwise).
- Looks good on FF and Chrome although interestingly the popup is 620 on Firefox, not 621

Selection_349
Selection_350


jacogr (approved):
Thank you!

PR #546, opened 2020-11-25
Set root height with vh instead of %

Tbaut:
I'm hopeful, looks good on chrome here at least :)


jacogr (approved):
Thank you! Let's try. It does make sense using vh here when a height is not specified.

Issue #538, opened 2020-11-20
Support Ethereum account


xlc:
https://github.com/polkadot-js/common/blob/6a8344020f33cf2b1c2d862fa7e5dfdf3da0e84d/packages/util-crypto/src/types.ts#L14

There is ethereum account support on keyring, but the exported JSON cannot be imported into extension

jacogr:
ATM, the fact that the JSON cannot be imported is actually good, I've been meaning to check. The signing atm will do weird things with that pair type. (Apart from not being properly displayed).

xlc:
Do you plan to support generate Ethereum signature in near future (within 3 months time)?
We are happy to contribute if needed, but will need your help to make sure we are doing it in the right way.

jacogr:
I don't have plans, but since it all works nicely on Moonbeam in the apps UI would certainly appreciate PRs to extend support here as well.

I took a quick look, so the signing itself should just work - https://github.com/polkadot-js/extension/blob/master/packages/extension-base/src/background/RequestExtrinsicSign.ts#L17-L24 - it works in the API and apps UI, so should do the same here now since the util support is there.

So it will probably be display only -

- on import, adjust the verify logic for ethereum accounts (called here https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/Popup/RestoreJson.tsx#L33 done here https://github.com/polkadot-js/extension/blob/master/packages/extension-base/src/background/handlers/Extension.ts#L262 - the apps UI has a similar thing I believe on create)
- default the IdentityIcon to ethereum fo these accounts - https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/components/Address.tsx#L139
- adjust re-encoding for these (no ss58) - https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/components/Address.tsx#L59
- not sure about the injection if magic is applied there....

So atm, think it is all cosmetic. Users of the extension may have some tweaks as well, i.e. for instance I think the apps UI may not quite handle these properly when injected on non-ETH chains (will have to see/test)

Tbaut:
Happy to have a look at this in the next days.

PR #554, opened 2020-12-01
Refactor JSON restoration

Tbaut:
- Use info coming from `KeyPair` (from the background script) rather than anything the file contains to be future proof, linked to #538.
- Show error on wrong file
- Show error on wrong password

restore


jacogr (approved):
Looks sane from a code pov. 

Have not played though, but knowing you it does what it says on the tin.

PR #560, opened 2020-12-02
Support ethereum accounts

### LINKED ISSUE #538: Support Ethereum account

Tbaut:
closes #538

- Restore with JSON
- Export JSON
- Forget
- Forbid derivation

ethereum


jacogr:
That `KeyringJson`, as injected, is much simpler. It basically has the name & address, no other info. Have not looked at your hack....

Tbaut:
Nice, I'll PR that to ui-keyring now.

Tbaut:
I think it's good to go for a first look @jacogr

The `Address` now accepts a new `type` prop. Based on this the address will be displayed with the ethereum encoding. If this is type is not supplied (in most cases) we'll check if we have this account in our Keyring.

There's a new checkbox both for account creation with mnemonic or when recovering an account with mnemonic since we have no way to know if it's an Ethereum account.

The accounts aren't injected in `apps`, you may know why.

jacogr:
I have been ignoring this one, will get in after the update on Monday.

Not sure why not injected, not a bad thing atm - I believe it is a keyring issue with the re-encoding. ... possibly. With that not-working, it still is a good idea to get it in, the issue is that we will probably break apps (apps === any user, not just the apps UI) when we start injecting these addresses.

Not sure how to solve that, may need to deprecate the current accounts interface (or add a flag to it), to have some additional info coming along. Not sure.

Tbaut:
Although the css for the image is a bit broad (img selector), it doesn't seem to break anything.
image

jacogr (approved):
Believe this is good to go - it works as-is, however users would need some additional effort (see the `get(true)` changes to actually start using these.

PR #561, opened 2020-12-04
leverage the phishing list caching - Move phishing check to background script

Tbaut:
related to https://github.com/polkadot-js/phishing/pull/26

- moved any phishing check to the background. One function will check if the url is part of the denied list (cached every hour), redirect if true.
- move phishing dependancy from the root `extension` to `packages/extension-base` where it's used
- remove the isXmlOrPdf util


packages/extension-base/src/page/index.ts
```diff
@@ -51,23 +50,16 @@ export function sendMessage<TMessageType extends MessageTypes> (message: TMessag
...
-    return new Injected(sendMessage);
-  }
+  return new Injected(sendMessage);
 }
 
-// the redirection function, called by the page in case the host
-// is flagged as a phishing site.
-export async function redirectPhishing (): Promise<Injected> {
-  await sendMessage('pub(phishing.redirect)');
+// check if the host of this page is part of the phishing list
```
Tbaut, line 58:
```suggestion
// redirect the use if this page is considered as phishing, otherwise return false
```

jacogr:
I still don’t get this xml or pdf thing ;)

Tbaut:
I'm still ¯\_(ツ)_/¯ so that's the right time to remove it then!

packages/extension-base/src/background/handlers/Tabs.ts
```diff
@@ -143,9 +144,21 @@ export default class Tabs {
     return null;
   }
 
+  private async redirectIfPhishing (url: string): Promise<boolean> {
+    const { deny } = await this.#state.getPhishingList();
+
+    const isInDenyList = checkHost(deny, url);
```
jacogr, line 150: Why not the `checkIfDeny` instead of re-doing the cache logic here?

Tbaut: fixed, that's a misunderstanding from my side :)

jacogr (approved):
lgtm. Hopefully it does what it says on the tin and skips some extra requests.

Tbaut:
According to my test, watching the network tab of the background process, there's is a call to this `all.json` at first page load, no more then when refreshing the page.

Issue #555, opened 2020-12-01
Misplaced error message


Tbaut:
from @tomusdrw 

![image](https://user-images.githubusercontent.com/33178835/100733182-18f63400-33ce-11eb-8a8e-2eb90dab069b.png)

PR #565, opened 2020-12-07
Fix signing tests and unskip them

### LINKED ISSUE #558: Fix and let Signing tests run
They are skipped right now https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/Popup/Signing/Signing.test.tsx#L27

Tbaut:
closes #558
- adds a test for wrong password error message

Todo:
- [x] make sure the extrinsics are decoded
- [ ] if possible add a test to prevent https://github.com/polkadot-js/extension/issues/555


packages/extension-ui/src/Popup/Signing/Signing.test.tsx
```diff
@@ -182,46 +228,70 @@ describe.skip('Signing requests', () => {
...
-  "value": 10000000000
-}`);
+      wrapper.find('FontAwesomeIcon.arrowRight').simulate('click');
+      expect(wrapper.find(Address).find('.fullAddress').text()).toBe('5E9nq1yGJJFiP8C75ryD9J2R62q2cesz6NumLnuXRgmuN5DG');
+      expect(wrapper.find(Extrinsic).find('td.data').map((el): string => el.text())).toEqual([
+        'https://polkadot.js.org/apps/?rpc=wss%3A%2F%2Fwestend-rpc.polkadot.io#/accounts',
+        '0xe143f23803ac50e8f6f8e62695d1ce9e4e1d68aa36c1cd2cfd15340213f3423e',
+        '45',
+        '3',
+        '0x0403c6111b239376e5e8b983dc2d2459cbb6caed64cc1d21723973d061ae0861ef690b00b04e2bde6f',
```
Tbaut, line 250: I might need a tip here @jacogr. This extrinsic (a transfer) isn't decoded in the test. Do you know a way for me load the latest metadata?

Tbaut:
>  if possible add a test to prevent https://github.com/polkadot-js/extension/issues/555

This is possible, but would blow up this suite of tests. I'll consider styling test with #577 

This is good to go, thanks a lot Jaco for your help on the decoding part :pray:

packages/extension-chains/src/index.ts
```diff
@@ -6,6 +6,7 @@ import type { Chain } from './types';
 
 import { Metadata } from '@polkadot/metadata';
 import { TypeRegistry } from '@polkadot/types';
+import { base64Decode } from '@polkadot/util-crypto';
```
jacogr, line 9: So am going to bump util-crypto with single-instance detection on util-crypto as well as metadata (and wasm-crypto, but that is stubbed). This may be an issue after it with useless logs. (Not your issue, and not 100% sure, so just a reminder to myself to check - the issue is that a window object is shared, so it really is wait and see atm...)

packages/extension-ui/src/messaging.ts
```diff
@@ -125,16 +125,17 @@ export async function createSeed (length?: SeedLengths, type?: KeypairType): Pro
   return sendMessage('pri(seed.create)', { length, type });
 }
 
+// eslint-disable-next-line @typescript-eslint/require-await
```
jacogr, line 128: This should not be needed? Await on line 141?

Tbaut: absolutely, that's a stray line from tests I did. Removing it.

jacogr (approved):

packages/extension-ui/src/messaging.ts
```diff
@@ -13,6 +12,8 @@ import { metadataExpand } from '@polkadot/extension-chains';
 import allChains from '@polkadot/extension-chains/chains';
 import chrome from '@polkadot/extension-inject/chrome';
 
+import { getSavedMeta, setSavedMeta } from './MetadataCache';
```
Tbaut, line 15: This is not particularly nice, but it appears that I can't use a `spyOn` test on a function that is not imported. So I had to move the `MetadataCache` out of `messagine`, to import them and be able to spy on them.

Also this `MetadataCache` might be better place somewhere else? I thought about a `./util` but it didn't feel right either.

PR #566, opened 2020-12-08
Add metamask extension web3

joelamouche:
- Create injectMetaMaskWeb3 function to inject MetaMask accounts and allow signing Tx
- First sends a request to MM to ask the user to authorize the app
- Metamask accounts are now displayed in the accounts section
- User can use theses accounts to sign tx
- Signing will trigger a MM window with a warning (because signing unprefixed messages is dangerous)


packages/extension-dapp/src/compat/web3Source.ts
```diff
@@ -90,7 +90,12 @@ function injectWeb3(win: Web3Window): void {
             },
           },
           signer: {
+            signPayload:async (payload: SignerPayloadJSON) : Promise<SignerResult>=>{
+              console.log('signPayload')
+              return {id:0,signature:await win.web3.eth.sign(JSON.stringify(payload), payload.address)}
```
jacogr:
```suggestion
              const raw = regsirty.createPayload('ExtrinsicPayload', payload).toU8a();
              
              return {id:0,signature:await win.web3.eth.sign(raw), payload.address)}
```

Pass the raw data, exactly as to be signed - not sure about `registry` here, not sure what is available in this context.

jacogr: PS; Sam with raw, you probably only need Raw - Metamask is never going to add decoding, so the logic in the API is -

- if signPayload, use that
- if sigRaw, use that
- otherwise, bail

So raw ill just be the actual raw data and cleans it up nicely.

packages/extension-dapp/src/compat/web3Source.ts
```diff
@@ -0,0 +1,137 @@
...
+            });
+
+            return (): void => {
+              // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+              sub.unsubscribe();
+            };
+          }
+        },
+        signer: {
+          signPayload: async (payload: SignerPayloadJSON) : Promise<SignerResult> => {
```
joelamouche, line 100: @jacogr could you take a look at this implementation of signPayload pls. I'm not really used to polkadot transactions yet. What is the signer supposed to sign? The function is signing a stringified paylaod but it doesn't seem to work. Thanks in advance

joelamouche: lol okay so you anticipated my question

jacogr: :D

packages/extension-dapp/src/compat/web3Source.ts
joelamouche: Im not sure I understand what I should do for the signRaw function

jacogr: It seems fine as-is. Remove `signPayload`, only keep that one.

joelamouche: but it seems signPayload is the one being called for transfers

jacogr: If available, like I mentioned, the order is -

- use signPayload, if available, other ...
- use signRaw, if available, otherwise ..
- ... bail

https://github.com/polkadot-js/api/blob/master/packages/api/src/submittable/createClass.ts#L292-L298

joelamouche:
updated with current state of my modifs

transfer get stuck here:
Screenshot 2020-12-10 at 17 05 03

packages/extension-dapp/src/compat/web3Source.ts
joelamouche: ok and should I sign raw.data, registry.createType('ExtrinsicPayload', raw) or registry.createType('ExtrinsicPayload', raw.data);

jacogr: Just the `raw.data` in `signRaw` is 100% - it is the actual data.

joelamouche: I updated back to using only signRaw and it gets stuck at the 'signing' step that I screenshot below. Let me know if it works for you

jacogr: Does this get logged?

https://github.com/polkadot-js/extension/pull/566/files#diff-85bb08733e1b7975e81532efecc738a9900827f7c54c78069e49d4c0de1a48e8R77

Just display the address and payload there as well - worried that is is an apps-specific issue.

joelamouche: Yes it does get log

```
console.log('signraw',raw.data,raw.address)
            console.log('signature', await win.web3.eth.sign(raw.data, raw.address));
```
logs this
Screenshot 2020-12-10 at 17 27 07

joelamouche: so it seems the signature is never resolved by MetaMask

jacogr: Ok, that data looks correct. So it seems MM doesn't come back. May be useful having a try/catch around the `await` to see if anything gets thrown. It may not like raw hex.

joelamouche: good call. It doesnt log anything unfortunately. just realized that web3.eth.sign is sync. But wrapping the callback in a promise doesn't return anything either. Might be a problem with Metamask... They have been warning against eth_sign for a while now because it is a security issue

joelamouche: also it looks like the message would be appended with "\x19Ethereum Signed Message:\n" https://eth.wiki/json-rpc/API#eth_sign

jacogr: Ahhh, forgot about the below, even though it is something in tests at least in 2 places. 

That would be an issue.

joelamouche: Yes indeed. Another approach would be use the bi-compatibility of Moonbeam to post an EVM compatible tx through subtrate. Let me check with the team and get back to you

joelamouche:
MetaMask integration is back 🥳 

After a long inquiry, I discovered that eth_sign (unprefixed) is available.
Here is a working test: https://github.com/joelamouche/test-dapp/tree/testEthSign

joelamouche:
@jacogr There is a problem though: when I run `yarn polkadot-dev-copy-to apps` in extensions, and then `yarn start` in apps, I get:

```
ERROR in ../../node_modules/@polkadot/extension-dapp/compat/Web3Source.js 7:0-24
<e> Module not found: Error: Can't resolve 'web3' in '/Users/antoineestienne/Github_Files/apps/node_modules/@polkadot/extension-dapp/compat'
<e> resolve 'web3' in '/Users/antoineestienne/Github_Files/apps/node_modules/@polkadot/extension-dapp/compat'
<e>   Parsed request is a module
<e>   using description file: /Users/antoineestienne/Github_Files/apps/node_modules/@polkadot/extension-dapp/package.json (relative path: ./compat)
<e>     Field 'browser' doesn't contain a valid alias configuration
<e>     resolve as module
<e>       /Users/antoineestienne/Github_Files/apps/node_modules/@polkadot/extension-dapp/compat/node_modules doesn't exist or is not a directory
<e>       /Users/antoineestienne/Github_Files/apps/node_modules/@polkadot/extension-dapp/node_modules doesn't exist or is not a directory
<e>       /Users/antoineestienne/Github_Files/apps/node_modules/@polkadot/node_modules doesn't exist or is not a directory
<e>       /Users/antoineestienne/Github_Files/apps/node_modules/node_modules doesn't exist or is not a directory
<e>       looking for modules in /Users/antoineestienne/Github_Files/apps/node_modules
<e>         single file module
<e>           using description file: /Users/antoineestienne/Github_Files/apps/package.json (relative path: ./node_modules/web3)
<e>             no extension
<e>               Field 'browser' doesn't contain a valid alias configuration
<e>               /Users/antoineestienne/Github_Files/apps/node_modules/web3 doesn't exist
<e>             .js
<e>               Field 'browser' doesn't contain a valid alias configuration
<e>               /Users/antoineestienne/Github_Files/apps/node_modules/web3.js doesn't exist
<e>             .jsx
<e>               Field 'browser' doesn't contain a valid alias configuration
<e>               /Users/antoineestienne/Github_Files/apps/node_modules/web3.jsx doesn't exist
<e>             .mjs
<e>               Field 'browser' doesn't contain a valid alias configuration
<e>               /Users/antoineestienne/Github_Files/apps/node_modules/web3.mjs doesn't exist
<e>             .ts
<e>               Field 'browser' doesn't contain a valid alias configuration
<e>               /Users/antoineestienne/Github_Files/apps/node_modules/web3.ts doesn't exist
<e>             .tsx
<e>               Field 'browser' doesn't contain a valid alias configuration
<e>               /Users/antoineestienne/Github_Files/apps/node_modules/web3.tsx doesn't exist
<e>         /Users/antoineestienne/Github_Files/apps/node_modules/web3 doesn't exist
<e>       /Users/antoineestienne/Github_Files/node_modules doesn't exist or is not a directory
<e>       /Users/antoineestienne/node_modules doesn't exist or is not a directory
<e>       /Users/node_modules doesn't exist or is not a directory
<e>       /node_modules doesn't exist or is not a directory
<e>  @ ../../node_modules/@polkadot/extension-dapp/compat/index.js 4:0-41 8:38-48
<e>  @ ../../node_modules/@polkadot/extension-dapp/index.js 11:0-43 71:49-59
```

joelamouche:
And when I install `yarn add web3` I get:

```
ERROR in ../../node_modules/ethereumjs-util/dist/account.js 4:13-30
<e> Module not found: Error: Can't resolve 'assert' in '/Users/antoineestienne/Github_Files/apps/node_modules/ethereumjs-util/dist'
<e> 
<e> BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
<e> This is no longer the case. Verify if you need this module and configure a polyfill for it.
<e> 
<e> If you want to include a polyfill, you need to:
<e> 	- add a fallback 'resolve.fallback: { "assert": require.resolve("assert/") }'
<e> 	- install 'assert'
<e> If you don't want to include a polyfill, you can use an empty module like this:
<e> 	resolve.fallback: { "assert": false }
<e>  @ ../../node_modules/ethereumjs-util/dist/index.js 29:13-33
<e>  @ ../../node_modules/ethereumjs-tx/dist/transaction.js 14:24-50
<e>  @ ../../node_modules/ethereumjs-tx/dist/index.js 3:20-44
<e>  @ ../../node_modules/web3-eth-accounts/lib/index.js 35:18-54
<e>  @ ../../node_modules/web3-eth/lib/index.js 34:15-43
<e>  @ ../../node_modules/web3/lib/index.js 30:10-29
<e>  @ ../../node_modules/@polkadot/extension-dapp/compat/Web3Source.js 7:0-24 33:21-25
<e>  @ ../../node_modules/@polkadot/extension-dapp/compat/index.js 4:0-41 8:38-48
<e>  @ ../../node_modules/@polkadot/extension-dapp/index.js 11:0-43 71:49-59
<e>  @ ../react-api/src/Api.tsx 14:0-68 52:27-39 188:30-40
<e>  @ ../react-api/src/index.ts 3:0-65 6:0-124 6:0-124 6:0-124 6:0-124
<e>  @ ./src/Root.tsx 6:0-42 35:36-39
<e>  @ ./src/index.tsx 9:0-26 18:35-39
```

And similar errors for http, https, os

using yarn add to add them doesnt work

joelamouche:
Adapting the apps helped: https://github.com/polkadot-js/apps/pull/5216

The console shows the right logs and MetaMask does ask me for user permissions. but then the address doesn't display in the accounts. Can you refresh my memory: what do I need to do to display it and test it?

joelamouche:
I think the problem is that the injcted accounts are injected without a type and default to substrate type addresses

joelamouche:
https://github.com/polkadot-js/common/pull/969
https://github.com/polkadot-js/extension/pull/566
https://github.com/polkadot-js/ui/pull/464
https://github.com/polkadot-js/apps/pull/5216 (wip)

Are all part of a first milestone: displaying injected apps from metamask into the app

joelamouche:
@ubaistp ?

joelamouche:
@jacogr Great news! I was able to sign a simple transfer on the app with MetaMask. See the updated PR description

packages/extension-dapp/src/compat/metaMaskSource.ts
```diff
@@ -0,0 +1,103 @@
...
+            });
+            // TODO: add onchainchanged
+
+            return (): void => {
+              sub.unsubscribe();
+            };
+          }
+        },
+        signer: {
+          signRaw: async (raw: SignerPayloadRaw): Promise<SignerResult> => {
```
joelamouche, line 78: Here

hendis85 (approved):

packages/extension-dapp/src/index.ts
```diff
@@ -124,35 +142,43 @@ export async function web3Accounts ({ ss58Format }: Web3AccountsOptions = {}): P
...
+  );
 
   return accounts;
 }
 
-export async function web3AccountsSubscribe (cb: (accounts: InjectedAccountWithMeta[]) => void | Promise<void>, { ss58Format }: Web3AccountsOptions = {}): Promise<Unsubcall> {
+export async function web3AccountsSubscribe (
+  cb: (accounts: InjectedAccountWithMeta[]) => void | Promise<void>,
+  { ss58Format }: Web3AccountsOptions = {}
+): Promise<Unsubcall> {
```
jacogr, line 155: Change to the rest of this file seems unneeded (after spending some time to go through). Would suggest just keeping to the logic to make reviewing slightly easier. (Not a critisism, but beauty is in the eye of the beholder and since tastes differ, it does make it much harder to get to the actual meat than it should be here - if reformatting is needed, would suggest making a different PR which just focusses on that)

packages/extension-dapp/src/index.ts
```diff
@@ -63,35 +70,42 @@ export function web3Enable (originName: string): Promise<InjectedExtension[]> {
...
-
-                return (): void => {
-                  // no ubsubscribe needed, this is a single-shot
-                };
-              };
-            }
-
-            return { ...info, ...ext };
+  web3EnablePromise = documentReadyPromise(
+    (): Promise<InjectedExtension[]> =>
```
jacogr, line 74: Like the previous comment, quite difficult to extract readability improvements vs logic changes. It seems the same.

packages/extension-dapp/src/index.ts
```diff
@@ -46,13 +52,14 @@ export { isWeb3Injected, web3EnablePromise };
...
 function getWindowExtensions (originName: string): Promise<[InjectedExtensionInfo, Injected | void][]> {
   return Promise.all(
-    Object.entries(win.injectedWeb3).map(([name, { enable, version }]): Promise<[InjectedExtensionInfo, Injected | void]> =>
-      Promise.all([
-        Promise.resolve({ name, version }),
-        enable(originName).catch((error: Error): void => {
-          console.error(`Error initializing ${name}: ${error.message}`);
-        })
-      ])
+    Object.entries(win.injectedWeb3).map(
```
jacogr, line 55: No logic changes, only reformatting. In this case the polkadot-js style is actually very different. In these types of situations, the chain is actually split if needed. Would just suggest to revert.

jacogr (commented):
Would suggest to -

- remove the reformatting so it focussed on logic-only
- needs a flag to say "also inject ethereum-accounts" - it is only needed on a subset of chains, not all

packages/extension-dapp/src/compat/singleSource.ts
```diff
@@ -73,15 +74,16 @@ function injectSingleSource (win: SingleWindow): void {
 // https://github.com/cennznet/singlesource-extension/blob/f7cb35b54e820bf46339f6b88ffede1b8e140de0/react-example/src/App.js#L19
 export default function initSingleSource (): Promise<boolean> {
   return new Promise((resolve): void => {
-    window.addEventListener('load', (): void => {
-      const win = window as Window & SingleWindow;
+    // window.addEventListener('load', (): void => {
+    console.log('loading singlesource');
```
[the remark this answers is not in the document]
joelamouche: the logging or the whole singleSource file? (I think it was there as a placeholder)

packages/extension-dapp/src/index.ts
joelamouche: Is there an easy way to revert the formating? Shouldn't `yarn run lint` do the trick? Thats what I ran on this repo

packages/extension-dapp/src/index.ts
joelamouche: I agree but I don't know how to revert this formatting that comes from the linter

joelamouche: Oh I guess I could run this PR against a formatted version 👍

joelamouche:
New PR without formatting

packages/extension-inject/src/types.ts
```diff
@@ -25,6 +25,8 @@ export interface InjectedAccountWithMeta {
     name?: string;
     source: string;
   };
+  type: KeypairType;
+  signer: Signer;
```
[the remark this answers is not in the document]
joelamouche: Where should it be? Is returning it in `enable` enough?

packages/extension-dapp/src/index.ts
```diff
@@ -104,18 +118,22 @@ export async function web3Accounts ({ ss58Format }: Web3AccountsOptions = {}): P
...
-        const list = await accounts.get();
-
-        return mapAccounts(source, list, ss58Format);
-      } catch (error) {
-        // cannot handle this one
-        return [];
+    injected.map(
+      async ({ accounts, name: source, signer }): Promise<InjectedAccountWithMeta[]> => {
+        try {
+          const list = await accounts.get();
```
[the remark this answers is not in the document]
joelamouche: So where do you think the filter should be?
op1: filter here the accounts by account.type
opt2: pass a flag to initCompat and only call the relevant sources in compat/index

packages/extension-base/src/background/handlers/Tabs.ts
```diff
@@ -30,7 +30,7 @@ function transformAccounts (accounts: SubjectInfo, anyType = false): InjectedAcc
       address,
       genesisHash,
       name,
-      type
+      type: type || 'sr25519'
```
jacogr, line 33: I would rather leave this empty - for some we really don't know the types and this is probably incorrect. so for Ledger it is ed25519, for Signer it is sr25519, but neither is guaranteed. 

So it really should always be has something or we just don't know. (At least in the cases where it does matter, it is there, eg `ethereum`)

packages/extension-dapp/src/compat/index.ts
```diff
@@ -1,11 +1,13 @@
...
 // SPDX-License-Identifier: Apache-2.0
 
+import initMetaMaskSource from './metaMaskSource';
 import singleSource from './singleSource';
 
 // initialize all the compatibility engines
 export default function initCompat (): Promise<boolean> {
   return Promise.all([
-    singleSource()
+    singleSource(),
```
jacogr, line 10: Note-to-self: After this can drop the singleSource. It was always great as a reference, but now that we have a proper reference available here, not needed anymore.

packages/extension-dapp/src/index.ts
```diff
@@ -1,12 +1,25 @@
...
+  InjectedAccount,
+  InjectedAccountWithMeta,
+  InjectedExtension,
+  InjectedExtensionInfo,
+  InjectedProviderWithMeta,
+  InjectedWindow,
+  ProviderList,
+  Unsubcall,
+  Web3AccountsOptions,
+} from "@polkadot/extension-inject/types";
```
jacogr, line 15: Just run `yarn lint` on it to fix the imports, the `"` here is wrong.

packages/extension-inject/src/types.ts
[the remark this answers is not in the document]
jacogr: It gets returned where we actually map the accounts. So when we ask for the account (source-from-account) the signer is returned.

jacogr (commented):
Small suggestions. I think the logic flow here is great overall. Just merge in master, get rid of the undeeded formatting and we should be close to get it in.

Really want to play with this.

On a side note  - please accept my humble apologies for letting this hang around so long without a review. It really has not been forgotten about, not seen as not-important, but I literally have not had the hours in the day recently.

packages/extension-dapp/src/compat/singleSource.ts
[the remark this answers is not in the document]
jacogr: Logging itself.

packages/extension-inject/src/types.ts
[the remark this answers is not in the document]
joelamouche: ok so in the enable function

joelamouche:
>On a side note

Oh no problem I know these last month have been very busy with the crowdloan and for me to.

Thanks for taking the time to getting back to this. 

@jacogr updated!

packages/extension-dapp/src/compat/metaMaskSource.ts
```diff
@@ -0,0 +1,103 @@
...
+            const signature = (await provider.request({ method: 'eth_sign', params: [raw.address, Web3.utils.sha3(raw.data)] })as string);
+
+            return { id: 0, signature };
+          }
+        }
+      };
+    },
+    version: win.web3.version
```
joelamouche, line 86: @jacogr what is expected here? MetaMask doesnt inject web3 anymore so I'm trying to get rid of web3 all together

hendis85 (approved):

packages/extension-dapp/src/compat/metaMaskSource.ts
joelamouche: @jacogr this is not needed, is it?

joelamouche:
@jacogr ready for review (tested)

joelamouche:
@jacogr updated and ready for review

packages/extension-base/src/background/handlers/Tabs.ts
```diff
@@ -30,7 +30,7 @@ function transformAccounts (accounts: SubjectInfo, anyType = false): InjectedAcc
       address,
       genesisHash,
       name,
-      type
+      type: type || undefined
```
jacogr, line 33: Does it come through as `null` somewhere? (If undefined the original is fully equivalent)

packages/extension-dapp/src/compat/metaMaskSource.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// add a compat interface of metaMaskSource to window.injectedWeb3
+function injectMetaMaskWeb3 (win: Web3Window): void {
+  // decorate the compat interface
+  win.injectedWeb3.Web3Source = {
+    enable: async (): Promise<Injected> => {
+      // win.web3 = new Web3(win.ethereum);
+
+      const providerRaw: unknown = await detectEthereumProvider({ mustBeMetaMask: true });
```
jacogr, line 53:
```suggestion
      const providerRaw: unknown = await detectEthereumProvider({ mustBeMetaMask: true });
```

packages/extension-dapp/src/compat/metaMaskSource.ts
```diff
@@ -0,0 +1,100 @@
...
+            // TODO: add onchainchanged
+
+            return (): void => {
+              sub.unsubscribe();
+            };
+          }
+        },
+        signer: {
+          signRaw: async (raw: SignerPayloadRaw): Promise<SignerResult> => {
+            const signature = (await provider.request({ method: 'eth_sign', params: [raw.address, Web3.utils.sha3(raw.data)] })as string);
```
jacogr, line 76:
```suggestion
            const signature = (await provider.request({ method: 'eth_sign', params: [raw.address, Web3.utils.sha3(raw.data)] }) as string);
```

packages/extension-dapp/src/compat/singleSource.ts
```diff
@@ -73,15 +73,13 @@ function injectSingleSource (win: SingleWindow): void {
 // https://github.com/cennznet/singlesource-extension/blob/f7cb35b54e820bf46339f6b88ffede1b8e140de0/react-example/src/App.js#L19
 export default function initSingleSource (): Promise<boolean> {
   return new Promise((resolve): void => {
-    window.addEventListener('load', (): void => {
-      const win = window as Window & SingleWindow;
+    const win = window as Window & SingleWindow;
```
jacogr, line 76: Believe I commented elsewhere - the original is correct. (All ok, since this will be dropped anyway)

packages/extension-dapp/src/index.ts
jacogr: Still applicable, but will do a manual revert of this and the one above once in.

jacogr (commented):
Think we are close. Some small comments and updates. 

(Bit difficult wading through the reformats as well to ensure that the logic is stil in-tact)

jacogr:
Added a deps PR (web3 & detect-provider) in #794 just to eliminate conflicts.

packages/extension-dapp/src/index.ts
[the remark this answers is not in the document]
joelamouche: @jacogr let me know if the latest commit corresponds to what you were thinking

joelamouche:
@jacogr updated and ready for review

packages/extension-dapp/src/index.ts
```diff
@@ -110,16 +121,18 @@ export async function web3Accounts ({ ss58Format }: Web3AccountsOptions = {}): P
...
-        return mapAccounts(source, list, ss58Format);
-      } catch (error) {
-        // cannot handle this one
-        return [];
+    injected.map(
+      async ({ accounts, name: source }): Promise<InjectedAccountWithMeta[]> => {
+        try {
+          const list = await accounts.get();
+
+          return mapAccounts(source, list.filter((acc) => { return acc.type ? acc.type === accountType : true; }), ss58Format);
```
jacogr, line 129:
```suggestion
          return mapAccounts(source, list.filter((acc) => acc.type && accountType ? acc.type === accountType : true), ss58Format);
```

If not specified, return all. (NOTE: This keeps it backwards compatible, however... still means that everybody using this needs to adapt)

My only concern now is that it does not allow you to ask for ed25519 & sr25519....

Maybe the cleaner way would be one of -

- excludes as well as include, or
- pass an array of values instead of a single, or
- an explicit isEthereum flag

Thoughts?

joelamouche:
@jacogr please review when you have some time

jacogr:
On it, want to get it in ASAP. (And importantly, get a release out with this)

Could you just look at the yarn.lock conflicts?

jacogr (approved):
Thank you.

PR #570, opened 2020-12-09
Build with esm config

jacogr:
Not for merging (yet) -

(a) this esm alongside cjs output is experimental
(b) obviously will need testing

Atm, taking a look to see if there is an actual output impact... as it stands, cannot see much/any saved (apart from sanity), so Webpack does a decent job even on the existing cjs-only stuff.


tsconfig.json
```diff
@@ -9,16 +9,28 @@
     "baseUrl": ".",
     "paths": {
       "@polkadot/extension": ["packages/extension/src"],
+      "@polkadot/extension/esm": ["packages/extension/src"],
```
jacogr, line 12: Not really needed, but sanity-wise simpler if all the imports does the same thing.

PR #572, opened 2020-12-10
Add a flag to get any account type for InjectedAccounts interface

### LINKED ISSUE #569: Adjust accounts, with type
Change this interface https://github.com/polkadot-js/extension/blob/master/packages/extension-inject/src/types.ts#L27-L30 to cater for non-ss58 accounts.

Basically add a `getAll()`, marking `get` as deprecated. With `getAll`, return `[<sr25519 | ed25519 | ethereum | ...>, InjectedAcocunt][]`

Since the type can be derived anyway from on-chain data, this is non-exposing. Then the logic would be -

- when `get` is called, only return the ss58-accounts, i.e. don't break userspace
- when `getAll` is called, return, well, everything with type

Needed for -

- https://github.com/polkadot-js/ui/pull/413
- https://github.com/polkadot-js/extension/pull/566
- https://github.com/polkadot-js/extension/pull/560

The alternative would be the following -

- add a `type?: KeyringType` to the AccountInjected
- add a flag to `get(withType?: boolean)` which will populate the type (if not supplied, same non-eth filter as above)

This is probably preferable since it has less working-with-data changes on the users of the interface, however they still need to opt-in.

Tbaut:
closes #569

### DIFF
--- a/packages/extension-base/src/background/handlers/Tabs.ts
+++ b/packages/extension-base/src/background/handlers/Tabs.ts
@@ -5,9 +5,8 @@ import type { InjectedAccount, InjectedMetadataKnown, MetadataDef, ProviderMeta 
 import type { KeyringPair } from '@polkadot/keyring/types';
 import type { JsonRpcResponse } from '@polkadot/rpc-provider/types';
 import type { SignerPayloadJSON, SignerPayloadRaw } from '@polkadot/types/types';
-import type { SingleAddress, SubjectInfo } from '@polkadot/ui-keyring/observable/types';
-import type { KeypairType } from '@polkadot/util-crypto/types';
-import type { MessageTypes, RequestAuthorizeTab, RequestRpcSend, RequestRpcSubscribe, RequestRpcUnsubscribe, RequestTypes, ResponseRpcListProviders, ResponseSigning, ResponseTypes, SubscriptionMessageTypes } from '../types';
+import type { SubjectInfo } from '@polkadot/ui-keyring/observable/types';
+import type { MessageTypes, RequestAccountList, RequestAuthorizeTab, RequestRpcSend, RequestRpcSubscribe, RequestRpcUnsubscribe, RequestTypes, ResponseRpcListProviders, ResponseSigning, ResponseTypes, SubscriptionMessageTypes } from '../types';
 
 import { PHISHING_PAGE_REDIRECT } from '@polkadot/extension-base/defaults';
 import { checkIfDenied } from '@polkadot/phishing';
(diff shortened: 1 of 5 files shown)


packages/extension-base/src/background/handlers/Tabs.ts
```diff
@@ -19,16 +20,27 @@ import RequestExtrinsicSign from '../RequestExtrinsicSign';
...
+  return getSortedAccounts(accounts)
     .map(({ json: { address, meta: { genesisHash, name } } }): InjectedAccount => ({
       address, genesisHash, name
     }));
 }
 
+function transformAllAccounts (accounts: SubjectInfo): [KeypairType, InjectedAccount][] {
+  return getSortedAccounts(accounts)
+    .map(({ json: { address, meta: { genesisHash, name } }, type }) => ([
+      type as KeypairType, { address, genesisHash, name }
```
Tbaut, line 40: type can be undefined here, hence the casting, will this bite any time or do you think `SubjectInfo` will always have it?

Tbaut:
lint is complaining on singleSource https://github.com/polkadot-js/extension/blob/master/packages/extension-dapp/src/compat/singleSource.ts#L52
I'm missing some context to know if this is something I should fix now, and how.

jacogr:
It may be better as option 2 - i.e. a flag to `get()` - that way the undefined doesn't matter at all, it is just part of InjectedAccount.

Tbaut:
Much better indeed, also it prevents another message :)

packages/extension-base/src/background/handlers/Tabs.ts
```diff
@@ -20,27 +19,21 @@ import RequestExtrinsicSign from '../RequestExtrinsicSign';
...
     .filter(({ json: { meta: { isHidden } } }) => !isHidden)
-    .sort((a, b) => (a.json.meta.whenCreated || 0) - (b.json.meta.whenCreated || 0));
-}
-
-function transformAccounts (accounts: SubjectInfo): InjectedAccount[] {
-  return getSortedAccounts(accounts)
-    .map(({ json: { address, meta: { genesisHash, name } } }): InjectedAccount => ({
-      address, genesisHash, name
+    .sort((a, b) => (a.json.meta.whenCreated || 0) - (b.json.meta.whenCreated || 0))
+    .map(({ json: { address, meta: { genesisHash, name } }, type }): InjectedAccount => ({
```
jacogr, line 27: If !withTypes, we just need to filter ETH, i.e. that is probably a user that will break. (Think of polkaseembly when it starts receiving these, all hell will break loose... :))

jacogr:
I'm happy to whack that - it was mostly kept as an example and there is one now with the Metamask integration.

packages/extension-base/src/background/handlers/Tabs.ts
```diff
@@ -19,13 +19,17 @@ import RequestExtrinsicSign from '../RequestExtrinsicSign';
 import State from './State';
 import { createSubscription, unsubscribe } from './subscriptions';
 
-function transformAccounts (accounts: SubjectInfo): InjectedAccount[] {
+function transformAccounts (accounts: SubjectInfo, anyType = false): InjectedAccount[] {
   return Object
     .values(accounts)
     .filter(({ json: { meta: { isHidden } } }) => !isHidden)
+    .filter(({ type }) => anyType ? true : type && ['ed25519', 'sr25519', 'ecdsa'].includes(type))
```
Tbaut, line 26: kept it as a separate filter to make it easier to read, can also integrate it above.

Tbaut: Am I right here btw, `ecdsa` isn't SS58, but the interface is compatible is it?

jacogr: Should be fine... hopefully :)

jacogr (approved):
Looks good, believe this should do the trick...

PR #575, opened 2020-12-11
Use account info to retrieve genesisHash in forget or export

### LINKED ISSUE #574: Address forget format
When having eg. an address attached to Kusama, it displays nicely with the prefix 2 in the overview list. When I try to forget it, it displays with the default prefix, 42. 

Basically on the forget screen we also need to show the chain and use that info for display.

Same actually happens with export, it just ignores the chain.

Tbaut:
closes #574

I'll log an issue, it'd help a lot to have test on such complex and important components as `Address`

chain

### DIFF
--- a/packages/extension-ui/src/components/Address.tsx
+++ b/packages/extension-ui/src/components/Address.tsx
@@ -44,6 +44,7 @@ interface Props {
 interface Recoded {
   account: AccountJson | null;
   formatted: string | null;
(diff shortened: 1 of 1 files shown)


Tbaut:
this will conflict with #560 I guess, I'll fix after merging.

jacogr (approved):
Have not tested, but trust the screenshot - code looks 100% sane.

PR #580, opened 2020-12-11
Swap Webpack to esm builds

jacogr:
This is with all polkadot packages dual-publishing esm (in addition to cjs) as well, so all inputs within our control are esm. 

```
cjs: 1733395 Dec 11 13:58 master-build.zip (master)
esm: 1275907 Dec 11 19:22 master-build.zip (this one)
```

```
cjs: 1519340 Dec 11 19:32 background.js
esm: 1280324 Dec 11 19:22 background.js
```

```
cjs: 2833776 Dec 11 19:32 extension.js
esm: 1453710 Dec 11 19:22 extension.js
```

UI itself is a massive saving.


packages/extension/webpack.config.js
```diff
@@ -93,11 +93,12 @@ module.exports = {
...
-    alias: packages.reduce((alias, pkg) => {
-      alias[`@polkadot/${pkg}`] = path.resolve(__dirname, `../${pkg}/src`);
-
-      return alias;
-    }, {}),
+    alias: packages.reduce((alias, p) => ({
+      ...alias,
+      [`@polkadot/${p}`]: path.resolve(__dirname, `../${p}/src`)
+    }), {
+      'react/jsx-runtime': require.resolve('react/jsx-runtime')
```
jacogr, line 100: Next version of React will allow to get rid of this little hack - the commit was merged, should be in the next release. 

Tracked this from babel, to babel reverted, to react-as-logged-by-babel, to commit... 

However, came up with the suggested fix without having done that hunt, that only came afterwards, stumbled across the above links while seeing if there is a simpler/better way, however, the only suggestion I saw was for this alias version, as already done.

jacogr:
Additionally, split the extension UI from the rest and disabled WASM compat in the UI, this is the result -

```
1099581 Dec 14 01:03 packages/extension/build/background.js
1074356 Dec 14 01:03 packages/extension/build/extension.js
```

In this case it actually doesn't make sense having the WASM in there -

- it adds loading time to the UI (WASM parsing & init)
- it is actually not really used, i.e. all keyring goes via persistent background (some hashing _may_ take place, but it is generally light and always has JS fallbacks)

Overall, with a smaller size and faster startup, it actually improves the UI.

jacogr:
@Tbaut Would just like a sanity check again, happy as it is. I couldn't spot anything "off", creating, removing, sending...

Tbaut (approved):
Looks all good :+1: 
- account creation
- export
- import
- signing without metadata
- metadata update and correctly decode
- account visibility
- account linked to specific chain
- remember password
- renaming

PR #581, opened 2020-12-11
Hide too long names behind ellipsis

### LINKED ISSUE #578: Text overflows for long names
followup to https://github.com/polkadot-js/extension/issues/414
image

Tbaut:
closes #578 
name

### DIFF
--- a/packages/extension-ui/src/components/Address.tsx
+++ b/packages/extension-ui/src/components/Address.tsx
@@ -35,7 +35,7 @@ interface Props {
   genesisHash?: string | null;
   isExternal?: boolean | null;
   isHidden?: boolean;
-  name?: React.ReactNode | null;
+  name?: string | null;
   parentName?: string | null;
   suri?: string;
   toggleActions?: number;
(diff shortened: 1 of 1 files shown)


packages/extension-ui/src/components/Address.tsx
```diff
@@ -35,7 +35,7 @@ interface Props {
   genesisHash?: string | null;
   isExternal?: boolean | null;
   isHidden?: boolean;
-  name?: React.ReactNode | null;
```
Tbaut, line 38: Not sure why this was the a node, it didn't sound sane, and didn't break anything (TS would tell us)

jacogr: Probably a catch-all. But since it is quite controlled, shouldn’t make a difference here.

Tbaut: right, since I use it below as a string, and TS was complaining that I couldn't use a node, shrinking this down helped me a little.

jacogr (approved):
Cool. Lgtm.

Issue #583, opened 2020-12-14
Unclicking "derive from root", still shows password


jacogr:
On the account creation it should sisappear when creating from a new seed.

PR #586, opened 2020-12-14
Hide parent password/path when non-deriving

jacogr:
Closes https://github.com/polkadot-js/extension/issues/583


Tbaut (approved):
:+1: Thanks, I'll add a test to your PR.

Also I find this flow a little weird. since the "+" menu doesn't have too many entries, I think it might make sense to have 2 entries there, 1 for derive, the other for new accounts.

jacogr:
Agreed. Atm it is 2 extra clicks while the user knows what he wants to do.

PR #594, opened 2020-12-17
Show error upon wrong password on Export screen

Tbaut:
closes #593 
- add loader to "forget button" (it's too quick on my computer, I can't even see it)
- add error to export with wrong password
- add tests for the above
error


jacogr:
Forget is weird. I have had some slow ones (as per the issue I logged) and then faster.

jacogr (approved):

PR #595, opened 2020-12-18
Prevent eth accounts to be selected for derivation

Tbaut:
An oversight from #560
We were still able to select an eth account as parent, and the `master` account, which is the default one selected when using the general "derivation menu" defaulted to the first account in the hierarchy, which could have been an eth one.

### DIFF
--- a/packages/extension-ui/src/Popup/index.tsx
+++ b/packages/extension-ui/src/Popup/index.tsx
@@ -8,6 +8,7 @@ import React, { useEffect, useState } from 'react';
 import { Route, Switch } from 'react-router';
 
 import { PHISHING_PAGE_REDIRECT } from '@polkadot/extension-base/defaults';
+import { canDerive } from '@polkadot/extension-base/utils';
 import uiSettings from '@polkadot/ui-settings';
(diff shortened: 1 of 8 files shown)


packages/extension-ui/src/Popup/index.tsx
```diff
@@ -51,7 +51,7 @@ async function requestMediaAccess (cameraOn: boolean): Promise<boolean> {
 
 function initAccountContext (accounts: AccountJson[]): AccountsContext {
   const hierarchy = buildHierarchy(accounts);
-  const master = hierarchy.find((account) => !account.isExternal);
+  const master = hierarchy.find((account) => !account.isExternal && account.type !== 'ethereum');
```
Tbaut, line 54: Thinking about it, I'll rather go for `['ed25519', 'sr25519', 'ecdsa'].includes(type)` to be future proof like I did elsewhere

jacogr (approved):
lgtm. We will need support for derive at some point with the bip standards. (I have been kicking against it since it brings in more mostly-unneeded deps)

PR #599, opened 2020-12-23
Some more tests for the Address component

Tbaut:
we've now covered most use-cases I think with:
- showing info from metadata (happens on signing popup)
- showing parent info (on main account overview only)


jacogr (approved):
Nice.

PR #602, opened 2020-12-28
Rm frames (align with apps)

jacogr:


Tbaut:
I saw it's a change that's quite old in apps https://github.com/polkadot-js/apps/pull/3245 But what does it do/prevent exactly?

jacogr:
It helps with clickjacking. Somebody wraps the page in a frame with JS listeners and gets access to what happens inside..

It is not really required in all these static pages, but it doesn't hurt to be safe rather than sorry. (Generally you want the server to send the correct headers but don't have control over those so this is the second-best)

PR #609, opened 2021-01-04
Show account address and name when importing from QR

### LINKED ISSUE #607: Importing an external account doesn't show the address or name
How to reproduce:
Import from a Qr code > scan Qr code --> no address or name displayed

Tbaut:
closes #607 
import

### DIFF
--- a/packages/extension-ui/src/components/ValidatedInput.tsx
+++ b/packages/extension-ui/src/components/ValidatedInput.tsx
@@ -27,6 +27,12 @@ function ValidatedInput<T extends Record<string, unknown>> ({ className, compone
   const isMounted = useIsMounted();
 
   useEffect(() => {
+    if (defaultValue) {
+      setValue(defaultValue);
+    }
+  }, [defaultValue]);
+
+  useEffect(() => {
     // Do not show any error on first mount
     if (!isMounted) {
       return;
--- a/packages/extension-ui/src/partials/Name.tsx
+++ b/packages/extension-ui/src/partials/Name.tsx
@@ -14,15 +14,16 @@ interface Props {
   label?: string;
   onBlur?: () => void;
   onChange: (name: string | null) => void;
+  value?: string;
 }
 
-export default function Name ({ address, className, isFocused, label, onBlur, onChange }: Props): React.ReactElement<Props> {
+export default function Name ({ address, className, isFocused, label, onBlur, onChange, value }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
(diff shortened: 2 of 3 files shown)


packages/extension-ui/src/components/ValidatedInput.tsx
```diff
@@ -26,6 +26,12 @@ function ValidatedInput<T extends Record<string, unknown>> ({ className, compone
   const [validationResult, setValidationResult] = useState<Result<string>>(Result.ok(''));
   const isMounted = useIsMounted();
 
+  useEffect(() => {
```
Tbaut, line 29: To me, this `useEffect` shouldn't be needed.. because the `defaultValue` prop change should make the whole `ValidatedInput` component rerender. Yet, it doesn't.. and the input isn't populated correctly without this `useEffect`. Does this come from the props type extending a `Record<string, string>`?

packages/extension-ui/src/partials/Name.tsx
```diff
@@ -14,15 +14,16 @@ interface Props {
...
 
-export default function Name ({ address, className, isFocused, label, onBlur, onChange }: Props): React.ReactElement<Props> {
+export default function Name ({ address, className, isFocused, label, onBlur, onChange, value }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const { accounts } = useContext(AccountContext);
   const isNameValid = useMemo(() => isNotShorterThan(3, t<string>('Account name is too short')), [t]);
 
   const account = accounts.find((account) => account.address === address);
-  const startValue = account && account.name;
+  const startValue = value || account?.name;
```
jacogr, line 26: This is the issue for the comment above - the above component only sets the value once. Here it is not available on the first render, so it never gets used.

jacogr (approved):

packages/extension-ui/src/partials/Name.tsx
Tbaut: Not sure I understand. Here `value` comes from the props and every time this changes, the `Name` component rerenders, and `startValue` will get a new value.

If I remove the `useEffect` from `ValidatedInput` and log both `startValue` from `Name` and `defaultValue` from `ValidatedInput`, then I get:
```
startValue undefined
defaultValue undefined
startValue root
defaultValue root
```

the new value, with "root" does end up in the `ValidatedInput`, but it doesn't re-render :man_shrugging:

jacogr: No, it doesn't re-create, it just passes props to the component.

In that component you only set the initial state right on creation.

PR #611, opened 2021-01-05
Add tests for External import flow

Tbaut:
follow-up to #609


Tbaut:
I can't find a way to use a mocked `@polkadot/react-qr` only when I need it. It seems that `automock` is always on.

The following tests use `@polkadot/react-qr`
- `packages/extension-ui/src/Popup/Signing/Signing.test.tsx`
- `packages/extension-ui/src/Popup/ImportQr.test.tsx`

the first one, `Signging` which breaks in this PR, doens't require a mock,  whereas the second one `ImportQr`, added in this PR and working well should use a mock for `QrScanAddress`.

I've placed the mocked module next to `node_modules` in `__mocks__/@polkadot/react-qr.tsx`, it is picked up by both tests although it should only be used by `ImportQr`

jest.config is set to reset to:
```
automock: false, // that's the default
clearMocks: true, // Automatically clear mock calls and instances between every test
```
Any or both `jest.unmock` and `jest.dontMock`, don't work, the mock is always picked up. Even worse, it seems that the mock is used even when I don't use `jest.mock`.

This is mentioned for instance [here](https://github.com/facebook/jest/issues/2649#issuecomment-657927950).
It looks like `automock: false` is ignored...

packages/extension-ui/src/Popup/ImportQr.test.tsx
```diff
@@ -0,0 +1,95 @@
...
+  wrapper.find('input').first().simulate('change', { target: { value } });
+  await act(flushAllPromises);
+  wrapper.update();
+};
+
+describe('ImportQr component', () => {
+  let wrapper: ReactWrapper;
+
+  beforeEach(async () => {
+    jest.doMock('@polkadot/react-qr');
```
Tbaut, line 31: the mock is picked-up even when removing this line.

jacogr:
So in the `.spec.ts` itself, you can use override-require. That will replace it with whatever local interface you have - https://github.com/polkadot-js/wasm/blob/master/packages/wasm-crypto/test/alias-imports.js#L4

Tbaut:
thanks, that would be a workaround though, I'd prefer to use the built in jest function, and actually understand why things don't work as expected. I'll continue to dig.

Tbaut:
It was far from straightforward, and that's the best I managed to get. I needed to mock `QrScanAddress` to prevent a "Worker not defined" error and hack around an event emmition, because `simulate` only works with `shallow`, but not with `mount` which we're using (it was throwing "TypeError: ReactWrapper::simulate() event 'scan' does not exist").

jacogr (approved):

PR #617, opened 2021-01-08
Add Web3AccountsOptions to web3FromAddress function

yuzhiyou1990:


packages/extension-dapp/src/index.ts
```diff
@@ -178,12 +178,12 @@ export async function web3FromSource (source: string): Promise<InjectedExtension
...
 
 // find a specific provider based on an address
-export async function web3FromAddress (address: string): Promise<InjectedExtension> {
+export async function web3FromAddress (address: string, { ss58Format }: Web3AccountsOptions = {}): Promise<InjectedExtension> {
   if (!web3EnablePromise) {
     return throwError('web3FromAddress');
   }
 
-  const accounts = await web3Accounts();
+  const accounts = await web3Accounts({ ss58Format });
```
jacogr, line 186: Since the signer keypairs operate on the actual raw (non-ss58), the actual solution is actually the next line.

Instad of simply doing `account.address === address` it should -

- decode the address into hex
- do the compare agains the list in hex

That way it always works on the raw representation, no matter which ss58 is passed in.

jacogr: ... and instead of the `u8aToHex`, we can just decode and use `u8aEq(decodedAddress, decodeAddress(account.address))` (which is more optimal since there is no u8a -> hex involved)

yuzhiyou1990:
```javascript
const allAccounts = await web3Accounts({ ss58Format: 42 });
const injector = await web3FromAddress(allAccounts[0].address);
```
Error: web3FromAddress: Unable to find injected 525BJ6fSuXDiTTHH8ubfpEpZRR7FZgupSLy6d1hbZZ3K4ZxE"

jacogr: As above, the check on line 107 is the right place to adjust.

jacogr:
Happy to accept a PR like discussed above :) (Can just re-open and make the changes?)

yuzhiyou1990:
> Happy to accept a PR like discussed above :) (Can just re-open and make the changes?)

New changes submitted.

packages/extension-dapp/src/index.ts
```diff
@@ -184,7 +185,11 @@ export async function web3FromAddress (address: string): Promise<InjectedExtensi
   }
 
   const accounts = await web3Accounts();
-  const found = address && accounts.find((account): boolean => account.address === address);
+  let found: InjectedAccountWithMeta | undefined;
+
+  if (address) {
+    found = accounts.find((account): boolean => u8aEq(decodeAddress(account.address), decodeAddress(address)));
```
jacogr, line 191:
```suggestion
    const accountU8a = decodeAddress(address);
    
    found = accounts.find((account): boolean => u8aEq(decodeAddress(account.address), accountU8a));
```

The reason for the suggestion is that we don't need to decode this on each iteration. (We could use `addressEq` but opt for the lower-level version so we don't need to do the same work each iteration until found)

jacogr (commented):
Should just optimize slightly, but apart from that, it is great and should do the trick. Thank you very much.

jacogr (approved):
Thank you!

Issue #579, opened 2020-12-11
Add a derivation path field when importing an account from seed


Tbaut:
follow up to https://github.com/polkadot-js/extension/issues/414

We could have an "advanced options" drop down to allow users to enter a derivation path. This would make this hidden feature visible.

PR #619, opened 2021-01-11
Refactor seed import and add explicit derivation path field

Tbaut:
What this PR contains:
- `AccountNamePasswordCreation` component created from what was previously in `Popup/CreateAccountAccountName`. This is the component with the name, password and password repeat.
- Use this new component in "Derive", "Account creation" and "Account import from seed" screens
- Create tests for it (moved most tests from CreateAccount.test)
- Enhance Derive tests to test the actual creation
- Add derivation path field in account import from seed screen (closes https://github.com/polkadot-js/extension/issues/579), I still went for an "advanced" drop down because I figured that having an empty locked field like the derivation was unintuitive.

All the  "Derive", "Account creation" and "Account import from seed" flows are well tested (with #620)

https://user-images.githubusercontent.com/33178835/104341509-2be64300-54fa-11eb-92d5-73ad56788325.mp4


Tbaut:
Moving to WIP, I identified some other areas we can change while building the tests.

Tbaut:
I'll stop there because this is already too big, although a lot of the changes concerns tests. I've built #620 alongside so everything I touched here is well tested.

jacogr (approved):
Ok, on this one mostly trusting you - only did a brief look-through since, yes, it is a bit unwiledly.

Tbaut:
yes sorry about that, I didn't expect to fall in the rabbit hole there :)

PR #622, opened 2021-01-14
Prevent from typing a derivation with `///password` since it is ignored 

### LINKED ISSUE #616: Error out on ///password derivation on the derive screen
The title says it all, the keyring derivation ignores it unless it's for a seed/mnemonic

Tbaut:
closes #616 
Added some more tests with derivation lock.

### DIFF
--- a/packages/extension-ui/src/Popup/Derive/SelectParent.tsx
+++ b/packages/extension-ui/src/Popup/Derive/SelectParent.tsx
@@ -34,7 +34,7 @@ export default function SelectParent ({ className, isLocked, onDerivationConfirm
   const passwordInputRef = useRef<HTMLDivElement>(null);
 
   useEffect(() => {
-    // forbid the use of password since Keyring ignored it
+    // forbid the use of password since Keyring ignores it
     if (suriPath?.includes('///')) {
       setIsValidPath(false);
     }
--- a/packages/extension-ui/src/Popup/Derive/SelectParent.tsx
+++ b/packages/extension-ui/src/Popup/Derive/SelectParent.tsx
@@ -4,7 +4,6 @@
 import React, { useCallback, useContext, useEffect, useMemo, useRef, useState } from 'react';
 
 import { canDerive } from '@polkadot/extension-base/utils';
-import { assert } from '@polkadot/util';
 
 import { AccountContext, ActionContext, Address, ButtonArea, InputWithLabel, Label, NextStepButton, VerticalSpace, Warning } from '../../components';
 import useTranslation from '../../hooks/useTranslation';
@@ -21,6 +20,9 @@ interface Props {
   onDerivationConfirmed: (derivation: { account: { address: string; suri: string }; parentPassword: string }) => void;
 }
 
+// match any single slash
+const singleSlashRegex = /([^/]|^)\/([^/]|$)/;
+
 export default function SelectParent ({ className, isLocked, onDerivationConfirmed, parentAddress, parentGenesis }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const onAction = useContext(ActionContext);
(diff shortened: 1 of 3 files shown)


packages/extension-ui/src/Popup/Derive/SelectParent.tsx
```diff
@@ -30,8 +30,16 @@ export default function SelectParent ({ className, isLocked, onDerivationConfirm
   const [suriPath, setSuriPath] = useState<null | string>(defaultPath);
   const [parentPassword, setParentPassword] = useState<string>('');
   const [isProperParentPassword, setIsProperParentPassword] = useState(false);
+  const [isValidPath, setIsValidPath] = useState(true);
   const passwordInputRef = useRef<HTMLDivElement>(null);
 
+  useEffect(() => {
+    // forbid to use password
```
Tbaut, line 37:
```suggestion
    // forbid the use of password since Keyring ignored it
```

packages/extension-ui/src/Popup/Derive/SelectParent.tsx
```diff
@@ -30,8 +30,16 @@ export default function SelectParent ({ className, isLocked, onDerivationConfirm
   const [suriPath, setSuriPath] = useState<null | string>(defaultPath);
   const [parentPassword, setParentPassword] = useState<string>('');
   const [isProperParentPassword, setIsProperParentPassword] = useState(false);
+  const [isValidPath, setIsValidPath] = useState(true);
   const passwordInputRef = useRef<HTMLDivElement>(null);
 
+  useEffect(() => {
+    // forbid the use of password since Keyring ignored it
```
Tbaut, line 37:
```suggestion
    // forbid the use of password since Keyring ignores it
```

packages/extension-ui/src/Popup/Derive/SelectParent.tsx
```diff
@@ -131,17 +147,20 @@ export default function SelectParent ({ className, isLocked, onDerivationConfirm
...
-              {!suriPath && (
+              {(!suriPath || !isValidPath) && (
                 <Warning
                   isBelowInput
                   isDanger
                 >
-                  {t('Incorrect derivation path')}
+                  {!suriPath
+                    ? t('Incorrect derivation path')
+                    : t('`///password` not supported for derivation')
```
jacogr, line 162: Not 100% sure about this - there could be other errors? (I guess it is fine for a start)

Tbaut: The other option, I thought about was to throw in the background process, when validating the suri. I would have to change a little the message api and add a `allowPassword`, but I found that odd to have a request + answer message for something we know beforehand, and that is only relevant for this screen.

jacogr: Yes, also had that brief thought atm - but it adds a loooong delay in actually making the request.

jacogr: The apps UI has this validation function -

```js

function deriveValidate (suri: string, pairType: KeypairType): string | null {
  if (suri.includes('///')) {
    return 'Password paths are not supported on keys derived from others';
  }

  try {
    const { path } = keyExtractPath(suri);

    // we don't allow soft for ed25519
    if (pairType === 'ed25519' && path.some(({ isSoft }): boolean => isSoft)) {
      return 'Soft derivation paths are not allowed on ed25519';
    }
  } catch (error) {
    console.error(error);

    return (error as Error).message;
  }

  return null;
}
```

jacogr: pairType there is actually incorrect, it should be `pairType !== sr25519`... (and then with a relevant change to the error)

Tbaut: I remember checking it the other day. We also used to have the `keyExtractPath` in the UI. The reason I ditched it in the last PR is that it was a duplicate because also used in the background when creating an account.

I'll make the error handling a little more generic and add the check for sr25519.

jacogr (approved):
lgtm...

PR #624, opened 2021-01-15
Ledger nano support

### LINKED ISSUE #154: Add Ledger Nano S support
- First available hardware device (general hardware support asked for in #143) 

Integration not generally available as of yet, but dev builds are available for the first milestone - not perfect, but is at a point where it can be integrated and tested

cc @jleni

Tbaut:
closes #154

workflow and several expected errors displayed:
- expected error: First I select a network which I don't have the app for.
- expected error: I select Kusama, which is the app I have on the ledger, but my ledger is locked.
- I unlock it and refresh.

https://user-images.githubusercontent.com/33178835/104940289-d2768c00-59b1-11eb-968d-9620ba555e05.mp4

TODO:

- [x] Make sure it works when first loading the page (without genesis)
- [x] Make sure it reloads the page (and works) when changing the genesis
- [x] use usb icon for ledger accounts
- [x] persist the account across reloads see https://github.com/polkadot-js/ui/pull/431
- [x] create workflow for first time ledger users
- [x] test signing something


packages/extension-ui/src/Popup/ImportLedger.tsx
```diff
@@ -0,0 +1,160 @@
...
+    // I don't get how this is possible, but sometimes addIndex was a string...
+    getLedger(genesis).getAddress(false, Number(accountIndex), Number(addressOffset))
+      .then(({ address }) => {
+        setIsBusy(false);
+        setAddress(address);
+      }).catch((e: Error) => {
+        setIsBusy(false);
+        const errorMessage = e.message.includes('unkown')
+          ? t<string>(
+            'Is your ledger locked? Error from ledger: {{errorMessage}}',
```
jacogr, line 56: Ahhhh... ok, this is a simple solution for the apps UI - if Ledger, get the address first to ensure it is connected. then sign.

jacogr:
With the above comment, this is what we should just change - https://github.com/polkadot-js/ui/blob/master/packages/ui-keyring/src/observable/genericSubject.ts#L46

Tbaut:
I updated the first post with the current workflow.
Although it is not visible in the demo, the first time, users must open the extension in full screen in order to get the browser usb permission popup. This is not nice but I'll create a setup workflow to make this a little more seamless.

Tbaut:
The flow for signing.
- my ledger isn't unlocked (pin needs to be entered)
- the kusama app is not open
- I cancel the transaction on the ledger

https://user-images.githubusercontent.com/33178835/105181354-fcec5480-5b2b-11eb-8ea4-1879c8b7d4d8.mp4

Tbaut:
I'm ready for a first review.

Here's the first time flow:
- it shows "Connect to ledger" in the menu and opens a new tab if ledger was never linked
- expected error: My ledger was not unlocked, and didn't show up right away in the browser popup
- expected error: I selected Polkadot network although I only have Kusama App on my ledger
- expected error: I didn't open the app on the ledger at first

Once the first flow is over, we can add ledger accounts without needing to open the app in a new tab (not in the demo)

https://user-images.githubusercontent.com/33178835/105201558-527f2c00-5b41-11eb-97d4-d12ba38f4355.mp4

packages/extension-ui/src/hooks/useLedger.ts
```diff
@@ -0,0 +1,133 @@
...
+    }
+
+    setIsLoading(true);
+    setError(null);
+    setWarning(null);
+
+    // sometimes, somehow addIndex is perceived as a string
+    ledger.getAddress(false, Number(accountIndex), Number(addressOffset))
+      .then((res) => {
+        console.log('restult', res);
```
Tbaut, line 100:
```suggestion
```

packages/extension-ui/src/hooks/useLedger.ts
```diff
@@ -0,0 +1,133 @@
...
+        setIsLocked(true);
+        setWarning(warningMessage);
+        setError(t<string>(
+          'Ledger error: {{errorMessage}}',
+          { replace: { errorMessage } }
+        ));
+        console.error(e);
+        setAddress(null);
+      });
+  // eslint-disable-next-line react-hooks/exhaustive-deps
```
Tbaut, line 123: This dep array should contain `t`, the translation function. But if it has it it, it triggers a useless re-rendering of this component, it's visible at ~0:28 in the demo [here](https://github.com/polkadot-js/extension/pull/624#issuecomment-763747721) when taking a look at the button loading status. I can't pin point why. Ever had such a problem?

jacogr: Weird. t should be relatively "stable".

Tbaut: Super weird, I've tried many things like removing the `{ replace : {}}`, take the `t` to another function with `useCallback`, it keeps re-rendering whatever I put it in. And because `useTranslation` is in a hook, I can't memoize it. Maybe it's a side effect of `web2usb`? it refreshes when the `ledger` gets available.

Even weirder, it didn't happen when it wasn't in a custom hook, with the same code [here](https://github.com/polkadot-js/extension/blob/fb0b7671642b44c52f302496090f8372dba5822a/packages/extension-ui/src/Popup/ImportLedger.tsx#L96).

jacogr: Just add a note before that comment as to "why". (We will wonder a month from now)

However, leaving out t should be very safe I believe. So should not have any side-effects.

packages/extension-ui/src/Popup/Signing/LedgerSign.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }, [refresh, setError]);
+
+  const _onSignLedger = useCallback((): void => {
+    if (!ledger || !payload || !onSignature) {
+      return;
+    }
+
+    setError(null);
+    setIsBusy(true);
+    ledger.sign(payload.toU8a(true), Number(accountIndex), Number(addressOffset))
```
jacogr, line 49: This seriously worries me. This feels like a plaster on some issue, somewhere.

packages/extension-ui/src/Popup/Signing/LedgerSign.tsx
```diff
@@ -0,0 +1,104 @@
...
+      .then((signature) => {
+        onSignature(signature);
+      }).catch((e: Error) => {
+        setError(e.message);
+        setIsBusy(false);
+      });
+  },
+  [accountIndex, addressOffset, ledger, onSignature, payload, setError]
```
jacogr, line 57:
```suggestion
    },
    [accountIndex, addressOffset, ledger, onSignature, payload, setError]
```

packages/extension-ui/src/Popup/Signing/LedgerSign.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <div className={className}>
+      {!!ledgerWarning && (
+        <Warning>
+          {ledgerWarning}
+        </Warning>
+      )}
+      {error && (
+        <Warning
+          isDanger
+        >
```
jacogr, line 70:
```suggestion
        <Warning isDanger>
```

packages/extension-ui/src/Popup/Signing/LedgerSign.tsx
```diff
@@ -0,0 +1,104 @@
...
+        </Warning>
+      )}
+      {ledgerLocked
+        ? (<Button
+          isBusy={isBusy || ledgerLoading}
+          onClick={_onRefresh}
+        >
+          <FontAwesomeIcon
+            icon={faSync}
+          />
```
jacogr, line 81:
```suggestion
          <FontAwesomeIcon icon={faSync} />
```

Like above slightly less "visual noise".

packages/extension-ui/src/hooks/useLedger.ts
```diff
@@ -0,0 +1,133 @@
...
+function getNetwork (genesis: string): Network | undefined {
+  return ledgerChains.find(({ genesisHash }) => genesisHash[0] === genesis);
+}
+
+function retrieveLedger (genesis: string): Ledger {
+  let ledger: Ledger | null = null;
+
+  const { isLedgerCapable } = getState();
+
+  assert(isLedgerCapable, 'Incompatible browser, only Chrome is supporter');
```
jacogr, line 40:
```suggestion
  assert(isLedgerCapable, 'Incompatible browser, only Chrome is supported');
```

packages/extension-ui/src/hooks/useLedger.ts
```diff
@@ -0,0 +1,133 @@
...
+
+      return;
+    }
+
+    setIsLoading(true);
+    setError(null);
+    setWarning(null);
+
+    // sometimes, somehow addIndex is perceived as a string
+    ledger.getAddress(false, Number(accountIndex), Number(addressOffset))
```
jacogr, line 98: As per the above "worried" comment :)

packages/extension-ui/src/Popup/Signing/LedgerSign.tsx
Tbaut: It definitely is. I'll dig a bit. It happens since the first time I've played with the Dropdown. Getting an error for ledger, along the line "only integer are allowed".

packages/extension-ui/src/hooks/useLedger.ts
Tbaut: agreed, there's no way this can stay as is :)

packages/extension-ui/src/hooks/useLedger.ts
Tbaut:
```suggestion
  // If the dependency array is exhaustive, with t, the translation function, it
  // triggers a useless re-render when ledger device is connected.
  // eslint-disable-next-line react-hooks/exhaustive-deps
```

packages/extension-ui/src/Popup/Signing/LedgerSign.tsx
Tbaut: The string casting comes from the Dropdown, that is an enhanced HTML `select`. The "target.event.value" coming from the `onChange` is a string, this comes from the `HTMLSelectElement`. I've tried to extend this type to support `number` as a value, but failed so far. The best I could do is cast these right away in the onchange, because I know they are of type number.

jacogr: Ahhh. Ok, you may want to do the `Number(...)` in a useCallback then? I'm more worried that it propagates everywhere at this point - UI, back-end, may end up in the JSON, etc.

jacogr: ... which is indeed what you did :)

packages/extension-ui/src/Popup/ImportLedger.tsx
```diff
@@ -0,0 +1,177 @@
...
+      <VerticalSpace/>
+      <ButtonArea>
+        {ledgerLocked
+          ? (<Button
+            isBusy={ledgerLoading || isBusy}
+            onClick={refresh}
+          >
+            <FontAwesomeIcon
+              icon={faSync}
+            />
```
jacogr, line 156:
```suggestion
            <FontAwesomeIcon icon={faSync} />
```

packages/extension-ui/src/Popup/ImportLedger.tsx
```diff
@@ -0,0 +1,177 @@
...
+            {t<string>('Refresh')}
+          </Button>
+          )
+          : (<Button
+            isBusy={ledgerLoading || isBusy}
+            isDisabled={!!error || !!ledgerError || !address || !genesis}
+            onClick={_onSave}
+          >
+            {t<string>('Import Account')}
+          </Button>)
```
jacogr, line 166: This form (which you mostly use) feels weird to me, but maybe just because of what I'm used to. Since in this form you don't need the `()`'s actually. The work better in this form

```js
(
  <Button
    ...
  />
)
```

packages/extension-ui/src/Popup/Signing/LedgerSign.tsx
```diff
@@ -0,0 +1,101 @@
...
+  addressOffset? : number;
+  className?: string;
+  error: string | null;
+  genesisHash?: string;
+  onSignature?: ({ signature }: { signature: string }) => void;
+  payload?: ExtrinsicPayload;
+  setError: (value: string | null) => void;
+}
+
+function LedgerSign ({ accountIndex, addressOffset, className, error, genesisHash, onSignature, payload, setError } : Props): JSX.Element {
```
jacogr, line 26:
```suggestion
function LedgerSign ({ accountIndex, addressOffset, className, error, genesisHash, onSignature, payload, setError } : Props): React.ReactElement<Props> {
```

packages/extension-ui/src/partials/MenuAdd.tsx
```diff
@@ -72,12 +81,38 @@ function MenuAdd ({ className, reference }: Props): React.ReactElement<Props> {
...
+            : ''
+          }
           to='/account/import-qr'
         >
           <FontAwesomeIcon icon={faQrcode} />
           <span>{t<string>('Attach external QR-signer account')}</span>
         </Link>
       </MenuItem>
+      <MenuItem className='menuItem ledger'>
+        { isLedgerEnabled
```
jacogr, line 95:
```suggestion
        {isLedgerEnabled
```

jacogr (approved):
Small comments. Have not played through (have a Ledger needing setup at least now), so can do that once it is in.

packages/extension-ui/src/Popup/ImportLedger.tsx
Tbaut: I'm constantly on the fence with these, although I thought I was clear. it's like with 
```js
useCallback(() => {
},[ ]);
```

vs

```js
useCallback(
  () => {
  },
  [ ]
);
```

packages/extension-ui/src/Popup/ImportLedger.tsx
```diff
@@ -147,21 +147,24 @@ function ImportLedger ({ className }: Props): React.ReactElement {
...
       <ButtonArea>
         {ledgerLocked
-          ? (<Button
-            isBusy={ledgerLoading || isBusy}
-            onClick={refresh}
-          >
-            <FontAwesomeIcon icon={faSync} />
-            {t<string>('Refresh')}
-          </Button>
+          ? (
```
jacogr, line 150: Yes, I know it is more space-y ... beauty is in the eye of the beerholder I guess :)

PR #627, opened 2021-01-22
Add network selector when importing an account with mnemonic, and refactor

### LINKED ISSUE #457: Allow network selection when importing an account from mnemonic seed
Caused quite a bit of confusion when trying to transfer DOT to my wallet. Until I realised the default was a substrate address. I think if its a polkadot extension, it should default to the polkadot address.

Tbaut:
closes #457

Needs tests

https://user-images.githubusercontent.com/33178835/105499332-9a2cc180-5cc1-11eb-8c3e-8c974b0577a4.mp4


jacogr (approved):
Nice.

packages/extension-ui/src/Popup/ImportSeed/SeedAndPath.tsx
```diff
@@ -45,7 +45,7 @@ function SeedAndPath ({ className, onAccountChange, onNextStep }: Props): React.
     validateSeed(suri).then((newAccount) => {
       setError('');
       setAddress(newAccount.address);
-      onAccountChange({ ...newAccount });
+      onAccountChange(newAccount);
```
Tbaut: What I would need here, is `{ ...newAccount, genesis }` to take the genesis into account. but for some reason this breaks the tests, even `{ ...newAccount}` breaks the tests.. although it's the same as `newAccount` which passes fine..

jacogr (approved):
Sorry about the delay here, I seem to have forgotten to click :(

Tbaut:
ah no no issue, you did click and I'm back at this test problem, really want to get to the bottom of it before merging.

jacogr:
Ahhh, now I see :)

packages/extension-ui/src/Popup/ImportSeed/SeedAndPath.tsx
Tbaut: Giving up on that using Object.assign instead of the spread operator.
It looks like we're using all the relevant `@babel/plugin-proposal-object-rest-spread` from https://github.com/polkadot-js/dev/blob/master/packages/dev/config/babel-plugins.cjs still, jest hangs, and doesn't even spit out errors.

PR #634, opened 2021-01-27
Fix renaming alignment and behavior

Tbaut:
- fix input position (see below screenshot)
- allow typing "Enter" key to validate
- fix error, which had no background and was difficult to read

previously
image

now

https://user-images.githubusercontent.com/33178835/106025403-009e4f00-60c9-11eb-844b-38af6e6f7f52.mp4


jacogr (approved):
lgtm. 

This is sorely missing on the apps UI, not even sure where to edit an account (don't think it is possible, I think I whacked it at some point since it was horrible - but no complaints thus far...)

Tbaut:
> but no complaints thus far...

It's all what matters in the end :)

PR #636, opened 2021-01-28
Fix errors or warnings appearing in tests

Tbaut:
- silence expected console errors
- make sure state changes are wrapped into `act`
- clean deprecated config (see below)

### DIFF
--- a/jest.config.js
+++ b/jest.config.js
@@ -4,7 +4,6 @@
 const config = require('@polkadot/dev/config/jest.cjs');
 
 module.exports = Object.assign({}, config, {
-  browser: true,
   moduleNameMapper: {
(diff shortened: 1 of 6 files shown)


jest.config.js
```diff
@@ -4,7 +4,6 @@
 const config = require('@polkadot/dev/config/jest.cjs');
 
 module.exports = Object.assign({}, config, {
-  browser: true,
```
Tbaut, line 7: This is deprecated. I figured it's actually not even needed.

Tbaut:
This PR only touches tests and has for benefit to reduce the amount of log spam when running the tests. I think it's harmless.

jacogr:
Agreed. Seem to have lost track of it along the way. Seen it with the 0.37 release, but also didn’t want to merge anything right at the end.

jacogr (approved):

PR #641, opened 2021-02-01
Icon consistency

Tbaut:
- Use the same derive icon everywhere
- Use a QR for Signer accounts

image


jacogr (approved):
I detect some OCD kicking in :)

PR #642, opened 2021-02-04
Better extrinsic overview

### LINKED ISSUE #628: Make simple actions such as sending funds easier to review
Today users have to click a dropdown to actually see the amount and the destination address for super simple extrinsics such as transfer. We should make this more obvious and easier to verify what users are signing, at least for things like
- transfer (with or without reaping)
- bonding / unbonding

Tbaut:
closes #628 

a couple questions, see comments on Github.

with AccountId
image

with many LookupSources, known and unknown
image

### DIFF
--- a/packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
@@ -27,7 +27,9 @@ interface Props {
   url: string;
 }
 
-const ADDRESS_TYPE = ['AccountId', 'LookupSource'];
+const ADDRESS_TYPE = ['AccountId', 'AuthorityId', 'LookupSource'] as const;
+
+type AddressType = typeof ADDRESS_TYPE[number];
 
 function displayDecodeVersion (message: string, chain: Chain, specVersion: BN): string {
   return `${message}: chain=${chain.name}, specVersion=${chain.specVersion.toString()} (request specVersion=${specVersion.toString()})`;
@@ -54,7 +56,18 @@ function decodeMethod (data: string, chain: Chain, specVersion: BN): Decoded {
   return { args, method };
 }
 
-function Method ({ data, decoded: { args, method }, t }: { data: string, decoded: Decoded, t: TFunction }): React.ReactElement {
+function showAddress (type: AddressType, data: unknown, chain: Chain): React.ReactElement {
+  const arg = chain.registry.createType(type, data);
+
+  console.log('arg as string', arg.toString());
+
+  return <Address
+    address={arg.toString()}
+    isSmall={true}
+  />;
+}
+
+function Method ({ chain, data, decoded: { args, method }, t }: { chain: Chain | null, data: string, decoded: Decoded, t: TFunction }): React.ReactElement {
   if (!args || !method) {
     return (
       <tr>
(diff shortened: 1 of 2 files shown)


packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
```diff
@@ -66,26 +68,38 @@ function renderMethod (data: string, { args, method }: Decoded, t: TFunction): R
...
+          {console.log(JSON.stringify(args, null, 2))}
+          {method.meta.args.map((arg, index) => (
+            <tr key={index}>
+              <td className='label'>{arg.name}</td>
+              <td className='data'>
+                {
+                  ADDRESS_TYPE.includes(arg.type.toHuman())
+                    ? (
+                      <Address
+                        address={args[index].Id || args[index]}
```
Tbaut, line 84: This is ugly.
`LookupSource` type is an object with the `Id` field containing the address, an `AccountId` type arg however has the address right away. What's your take on that, what's the right way to take those out?

Also TS is complaining:
> Element implicitly has an 'any' type because expression of type 'number' can't be used to index type 'string | number | true | AnyJson[] | { [index: string]: AnyJson; }'

Is casting to something like ` { [index: string]: AnyJson; }` the only way to go?

packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
```diff
@@ -66,26 +68,38 @@ function renderMethod (data: string, { args, method }: Decoded, t: TFunction): R
...
+                        isSmall={true}
+                      />
+                    )
+                    : <pre>{JSON.stringify(args[index], null, 2)}</pre>
```
Tbaut, line 88: TS is also complaining about this one in a similar fashion.
> Element implicitly has an 'any' type because expression of type 'number' can't be used to index type 'string | number | true | AnyJson[] | { [index: string]: AnyJson; }'.
  No index signature with a parameter of type 'number' was found on type 'string | number | true | AnyJson[] | { [index: string]: AnyJson; }'.
  
  Wondering if I should cast it or if there's something better to do.

packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
jacogr: LookupSource can be multiple things -

- a straight AccountId
- an IndicesLookup
- a MultiAddress

jacogr: So the above doesn't work at all, it only caters for 2 of those (not correctly) since it actually should detect the type and then apply as required.

packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
```diff
@@ -27,6 +27,8 @@ interface Props {
   url: string;
 }
 
+const ADDRESS_TYPE = ['AccountId', 'LookupSource'];
```
jacogr, line 30: Address, MultiAddress, IndicesLookup, EthereumAddress, ... (each with their own formats)

packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
Tbaut: Thanks, so do I understand correctly that the best way to do is to track down all the types that may have an address, and get the address out of each type accordingly? Is that even possible with `LookupSource` if you say what is behind may be some other type? How can I know what type is behind `LookupSource`?

jacogr: For instance in the case of MultiAddress (which is an enum) each entry behind it _is_ indeed a valid address. (So you cannot just check for Id)

The best way of getting the type and knowing how to handle it is actually via an instanceof check. Once this is in-place, you would handle each type. For instance on MultiAddress that will go live in Polkadot .28, the destination may be an AccountId or An AccountIndex, or a bare key.

Tbaut:
Closing this as I don't think I'll work on it further.

PR #649, opened 2021-02-09
Remove current tab and open landing page

Tbaut:
This behavior is a tiny bit better in that it shows the landing page first, because it gets opened in a new active tab. The phishing website is still there in another tab though.

I've experimented with `chrome.tabs.query` such as: 
```
 chrome.tabs.query({ currentWindow: true, url: phishingWebsite }, (res) => { console.log('res', res); })
```
but for some reason, the `res` is always empty if the website is opened in a new tab

### DIFF
--- a/packages/extension-base/src/background/handlers/Tabs.ts
+++ b/packages/extension-base/src/background/handlers/Tabs.ts
@@ -145,11 +145,12 @@ export default class Tabs {
     const encodedWebsite = encodeURIComponent(phishingWebsite);
     const url = `${chrome.extension.getURL('index.html')}#${PHISHING_PAGE_REDIRECT}/${encodedWebsite}`;
 
-    chrome.tabs.query({ active: true, currentWindow: true, windowType: 'normal' }, (tabs) => {
-      const currentTabId = tabs[0].id;
-
-      currentTabId && chrome.tabs.remove(currentTabId, () => {
-        chrome.tabs.create({ active: true, url });
+    chrome.tabs.query({ currentWindow: true }, (tabs) => {
+      tabs.forEach((tab) => {
+        if (tab.url === phishingWebsite) {
+          tab.id && chrome.tabs.remove(tab.id);
+          chrome.tabs.create({ active: true, url });
+        }
       });
     });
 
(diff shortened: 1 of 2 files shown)


packages/extension-base/src/background/handlers/Tabs.ts
```diff
@@ -145,7 +145,13 @@ export default class Tabs {
     const encodedWebsite = encodeURIComponent(phishingWebsite);
     const url = `${chrome.extension.getURL('index.html')}#${PHISHING_PAGE_REDIRECT}/${encodedWebsite}`;
 
-    chrome.tabs.update({ url });
+    chrome.tabs.query({ active: true, currentWindow: true, windowType: 'normal' }, (tabs) => {
```
jacogr, line 148: Shouldn' this be something like -

```js
chrome.tabs.query({ url: phishingWebsite }, (tabs) => {
  tabs.forEach((tab) => {
    chrome.tabs.update(tab.is, { url });
  });
});
```

So basically it checks all the available tabs for the url?

Tbaut: The `tabs` here is an empty array, so it ends up opening the new phishing tab in background, without showing the phishing landing page.

I've tried many things in the query, but I am never able to get this specific website.

Also it's actually hard to debug because the `tab` ojects from the `tabs` array do not have any `url`...

```
      chrome.tabs.query({ currentWindow: true }, (tabs) => {
        console.log('tabs', tabs);
        tabs.forEach((tab) => {
          console.log('url', tab.url);
          chrome.tabs.update({ url });
        });
      });
```
returns with 4 tabs + opening the phishing website in a new tab:
> 5 url undefined

^ this shows that we should be able to query the tab with phishing website, but this means we need its id... which I don't know how to get as of now.

jacogr: TL;DR PITA

Tbaut:
ah you were quicker, closing in favor of #650

jacogr:
Dunno why I didn't just push to this branch :(

PR #650, opened 2021-02-09
Phishing redirect, tab check

jacogr:
Replaced #649


Tbaut (approved):
working well even with query string, I guess the `query` is only taking the domain into account.

jacogr:
Good check, thank you. I only did check for different parts, not actual query strings.

PR #666, opened 2021-03-03
Batch export

BubbleBear:
# Description

Implement export all and batch import features.

# Preview


packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -255,7 +263,7 @@ export default class Extension {
 
   private jsonRestore ({ file, password }: RequestJsonRestore): void {
     try {
-      keyring.restoreAccount(file, password);
+      keyring.restoreAccount(file as any, password);
```
BubbleBear, line 266: TODO: remove.
added for suppressing TS error for partial test.

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -255,7 +263,7 @@ export default class Extension {
 
   private jsonRestore ({ file, password }: RequestJsonRestore): void {
     try {
-      keyring.restoreAccount(file, password);
+      keyring.restoreAccounts(file);
```
BubbleBear, line 266: TO BE DECIDED
it's currently password free.

packages/extension-ui/src/Popup/RestoreJson.tsx
```diff
@@ -47,28 +48,35 @@ function Upload ({ className }: Props): React.ReactElement {
...
+            });
         });
+      }
     }, []
   );
 
   const _onRestore = useCallback(
     (): void => {
-      if (!file || !password) {
+      if (!file) {
```
BubbleBear, line 79: TO BE DECIDED
password requirement is currently disabled.

chrisli30: I have tested it. I think the updated behavior aligns with our plan.

In single wallet Recovery, user is required to type in the password of the wallet for validation. However, in batch Recovery typing in all passwords is not practical, so it's better to skip the validation step here and leave it to the signing process.

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -95,6 +95,14 @@ export default class Extension {
     return { exportedJson: JSON.stringify(keyring.backupAccount(keyring.getPair(address), password)) };
   }
 
+  private async accountsBatchExport ({ addresses }: RequestAccountBatchExport): Promise<ResponseAccountExport> {
+    return {
+      exportedJson: JSON.stringify(
```
jacogr, line 100: As suggested on the ui repo, the ui-keyring should rather do the heavy-lifting. This means that wehn e.g. added in the apps UI, the logic is exactly the same. so move the `JSON.stringify` there.

BubbleBear: aye sir.

BubbleBear: My implementation doesn't add a new entry point for importing. It extends current `Restore account from backup JSON file`.
I suppose a new entry point would be confusing to users.

However, this implementation skips password confirming which used to be in importing process.
Thus I need your decision here, whether I should add a separate entry point for batch importing, or remove the password input from importing at all.

jacogr: Keep the password input. Keep as-is. We can just warp the keyring export with the new `jsonEncrypt` call and return an encypted value there. (Same for decrypt)

https://github.com/polkadot-js/common/blob/master/packages/util-crypto/src/json/encrypt.ts#L12

So for the contentType pass through `[batch-pkcs8]` (`pkcs8` is the account-only format)

(Data would be `stringToU8a(JSON.stringify(accounts))` in the keyring, decrypt would be the reverse)

jacogr: ^^^ all the above in the UI PR in the keyring, so it returns a JSON encoded or a JSON decoded.

BubbleBear: Encrypted JSON brings up some problems.

- Currently the uploaded file is parsed right away to fetch account info via `jsonGetAccountInfo`. It's impossible with encrypted file content.
- And it can't be decided if an uploaded file requires password unless it's sent to ui-keyring for recovery, and that's too late.

We need to think about it very carefully.

jacogr: It is the same as the normal JSON files. Some meta information is not encrypted. Like address, name, tags. In this case just follow the same approach.

BubbleBear: cool

BubbleBear:
@jacogr 
changed on request and updated preview gif, please review.

package.json
```diff
@@ -13,7 +13,7 @@
     "@polkadot/networks": "^6.0.1-5",
     "@polkadot/types": "^4.0.1-4",
     "@polkadot/util": "^6.0.1-5",
-    "@polkadot/util-crypto": "^6.0.1-5",
+    "@polkadot/util-crypto": "^6.0.2",
```
jacogr, line 16: Just remove this and merge master, it is on the latest 6.0.5

packages/extension-ui/package.json
```diff
@@ -20,7 +20,7 @@
     "@polkadot/react-identicon": "^0.70.2-2",
     "@polkadot/react-qr": "^0.70.2-2",
     "@polkadot/util": "^6.0.1-5",
-    "@polkadot/util-crypto": "^6.0.1-5",
+    "@polkadot/util-crypto": "6.0.2",
```
jacogr, line 23: Same as above, will currently conflict.

BubbleBear: conflicts resolved.

BubbleBear:
@jacogr hi sir.
it requires `ui-keyring` with `batch-export` code to pass linting. would you bump up dependency versions please?

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -261,6 +269,14 @@ export default class Extension {
     }
   }
 
+  private batchRestore ({ file }: RequestBatchRestore): void {
```
jacogr, line 272: We should pass the password through here as well.

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -95,6 +95,14 @@ export default class Extension {
     return { exportedJson: JSON.stringify(keyring.backupAccount(keyring.getPair(address), password)) };
   }
 
+  private async accountsBatchExport ({ addresses }: RequestAccountBatchExport): Promise<ResponseAccountExport> {
```
jacogr, line 98: Needs the password to encrypt the JSON export. (password-less it will be encoded with no encryption)

BubbleBear: ok, so this password is used to encrypt all accounts exported right? and we don't need to verify it.
I'll need to open another PR in `ui` repo.

BubbleBear: I find it a bit hard to design UI for entering password before exporting all.
Got any idea?

BubbleBear: The UI is rough, but the function solid.

jacogr (commented):
It looks really good. Will have a play-through in the morning, but don't see any immediate issues after looking through.

jacogr:
Happy to merge, would just need the linting fix. (`yarn lint` locally to fix what can be auto-fixed)

BubbleBear:
> Happy to merge, would just need the linting fix. (`yarn lint` locally to fix what can be auto-fixed)

ahhhh, this linting error is due to outdated `ui-keyring` package. it will be fixed after a dependency bump up.

jacogr (approved):
Thank you! This is really great.

PR #667, opened 2021-03-04
Support ipfs and ipns urls

carumusan:
Brave browser has native support for ipfs:// and ipns:// urls but this assertion is currently blocking. Updated to support them. Untested.


jacogr (approved):
Thank you.

Issue #631, opened 2021-01-25
Support signing payloads with custom signedExtensions


KarishmaBothara:
Support to sign transaction with userExtensions (signedExtension specific/custom to chains) .
Right now it injects the current signed extensions for encoding - https://github.com/polkadot-js/extension/blob/master/packages/extension-base/src/background/RequestExtrinsicSign.ts#L19

    ` registry.setSignedExtensions(this.payload.signedExtensions);
 `
Need support for 

    `registry.setSignedExtensions(this.payload.signedExtensions, userExtensions);`

Also on our chain we override the definition of `ChargeTransactionPayment`, however during signing it uses the default type definition

jacogr:
This is actually not solvable atm since the apps UI just doesn't know about these - and it is the injector for this information.

So to solve it requires a bit of work -

- the API typesBundle types needs to be adjusted to allow for signed extensions, specifically here https://github.com/polkadot-js/api/blob/master/packages/types/src/types/registry.ts#L47
- like it does with types, the API needs to expand the `userExtensions` alongside the types bundle and inject those for use, aka the same way as we do in the known types here for RPC https://github.com/polkadot-js/api/blob/master/packages/types-known/src/index.ts#L66

Once all that is done, the apps UI can be adjusted to inject the extensions as well, without that it will only always inject what is in the metadata, here https://github.com/polkadot-js/apps/blob/master/packages/page-settings/src/useChainInfo.ts#L26

After that, the extension can save this additionally and inject as required based on the chain.

This has not come up as a requirement, since the chains contributed in the apps UI above has not had that need (this is the provider of information). Sadly without that it is really working in a black box since there are no actual way of testing and getting it working end-to-end. 

TL;DR Certainly do-able, but not testable atm.

KarishmaBothara:
Thanks a lot for your reply and the detailed explanation...It is very helpful...
Just one doubt I have now is... If I don't use the apps UI and have UI in which I am trying to use the extension with....i.e using web3Enable, web3AccountsSubscribe to talk to extension... Once API(polkadot) is upgraded with the changes you pointed out.. how can my UI pass that information to extension?

jacogr:
The apps UI dictates the format at this point, so the extension doesn't know about it and would just ignore it completely. (As I mentioned, the apps UI is deemed as the primary provider of this information for extension upgrades and the 2 move in lock-step on the exchange formats)

[...]

PR #670, opened 2021-03-05
Added code to support custom signed extension/user extension

KarishmaBothara:
closes https://github.com/polkadot-js/extension/issues/631
Tests added.

### DIFF
--- a/packages/extension-base/src/background/RequestExtrinsicSign.ts
+++ b/packages/extension-base/src/background/RequestExtrinsicSign.ts
@@ -15,8 +15,6 @@ export default class RequestExtrinsicSign implements RequestSign {
   }
 
   sign (registry: TypeRegistry, pair: KeyringPair): { signature: string } {
-    // inject the current signed extensions for encoding
-    registry.setSignedExtensions(this.payload.signedExtensions);
 
     return registry
       .createType('ExtrinsicPayload', this.payload, { version: this.payload.version })
--- a/packages/extension-base/src/background/RequestExtrinsicSign.ts
+++ b/packages/extension-base/src/background/RequestExtrinsicSign.ts
@@ -15,7 +15,6 @@ export default class RequestExtrinsicSign implements RequestSign {
   }
 
   sign (registry: TypeRegistry, pair: KeyringPair): { signature: string } {
-
     return registry
       .createType('ExtrinsicPayload', this.payload, { version: this.payload.version })
       .sign(pair);
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -344,6 +344,7 @@ export default class Extension {
       // Get the metadata for the genesisHash
       const currentMetadata = this.#state.knownMetadata.find((meta: MetadataDef) =>
         meta.genesisHash === payload.genesisHash);
+
       // set the registry before calling the sign function
       registry.setSignedExtensions(payload.signedExtensions, currentMetadata?.userExtensions);
 
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -3,7 +3,7 @@
 
 import type { MetadataDef } from '@polkadot/extension-inject/types';
 import type { KeyringPair, KeyringPair$Json, KeyringPair$Meta } from '@polkadot/keyring/types';
-import type { SignerPayloadJSON } from '@polkadot/types/types';
+import type { SignerPayloadJSON, SignerPayloadRaw } from '@polkadot/types/types';
 import type { SubjectInfo } from '@polkadot/ui-keyring/observable/types';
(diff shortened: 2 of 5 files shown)


KarishmaBothara:
@jacogr - Request your review...

packages/extension-base/src/background/RequestExtrinsicSign.ts
```diff
@@ -15,9 +15,6 @@ export default class RequestExtrinsicSign implements RequestSign {
   }
 
   sign (registry: TypeRegistry, pair: KeyringPair): { signature: string } {
-    // inject the current signed extensions for encoding
```
jacogr, line 18: Not sure why this was removed? (As commented the signed extensions from the payload should be as-defined for the transaction)

packages/extension-chains/src/index.ts
```diff
@@ -40,10 +40,12 @@ export function metadataExpand (definition: MetadataDef, isPartial = false): Cha
   const isUnknown = genesisHash === '0x';
   let hasMetadata = false;
 
+  const signedExtensions = metaCalls ? new Metadata(registry, base64Decode(metaCalls)).asLatest.extrinsic.signedExtensions.toJSON() as string[] : undefined;
```
jacogr, line 43: Here we base64 decode the meta twice and create the Metadata twice. Should be done once only since it is quite a heavy operation, they are massive. (So move it inside the check below)

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -333,6 +334,19 @@ export default class Extension {
       pair.decodePkcs8(password);
     }
 
+    if (Object.prototype.hasOwnProperty.call(request.payload, 'genesisHash')) {
```
jacogr, line 337: The payload will always contain the genesisHash?

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -333,6 +334,19 @@ export default class Extension {
...
+      // Get the metadata for the genesisHash
+      const currentMetadata = this.#state.knownMetadata.find((meta: MetadataDef) =>
+        meta.genesisHash === (request.payload as SignerPayloadJSON).genesisHash);
+
+      // set the registry before calling the sign function
+      registry.setSignedExtensions((request.payload as SignerPayloadJSON).signedExtensions, currentMetadata?.userExtensions);
```
jacogr, line 343: Don't believe his makes sense - the payload will contain all the extensions as per the metadata, so these are known.

packages/extension-chains/src/index.ts
KarishmaBothara: thanks... made this change

packages/extension-base/src/background/handlers/Extension.ts
KarishmaBothara: There can be two different signing ways
**extrinsicSign** where payload is of type SignerPayloadJSON which contains the genesisHash
**bytesSign** (raw) where payload is of type SignerPayloadRaw which does not contain genesisHash (https://github.com/polkadot-js/api/blob/a9c9fb5769dec7ada8612d6068cf69de04aa15ed/packages/types/src/types/extrinsic.ts#L101)

packages/extension-base/src/background/RequestExtrinsicSign.ts
KarishmaBothara: It removed from here, and used in a level above this... i.e while handling signingApprovedPassword so the registry that's is passed will always have current signed extensions injected..
https://github.com/KarishmaBothara/extension/blob/dfdc1ed1eba183d9477d5c2c61ac63a707ff2a4f/packages/extension-base/src/background/handlers/Extension.ts#L343

packages/extension-base/src/background/handlers/Extension.ts
jacogr: Indeed. I misread the next call (I blame GH context here) as a direct call to the `sign` method.

So in this case, instead of casting create a `isJsonPayload (value: SignerPayloadJSON | SignerPayloadRaw): valus is SignerPayloadJSON` function and use in in the if

Also suggest just doing `!isUndefined((request.payload as SignerPayloadJSON).genesisHash` as the check there.

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -333,6 +338,19 @@ export default class Extension {
       pair.decodePkcs8(password);
     }
 
+    if (isJsonPayload(request.payload)) {
+      // Get the metadata for the genesisHash
+      const currentMetadata = this.#state.knownMetadata.find((meta: MetadataDef) =>
+        meta.genesisHash === (request.payload as SignerPayloadJSON).genesisHash);
```
jacogr, line 344:
```suggestion
        meta.genesisHash === request.payload.genesisHash;
```

The guard above ensures the value is what it is meant to be. (If TS does complain, which it shouldn't however since it is nsted, then `const { request } = payload` so it is stand-alone would nudge it in the right direction before the guard)

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -333,6 +338,19 @@ export default class Extension {
...
+
+      // set the registry before calling the sign function
+      registry.setSignedExtensions((request.payload).signedExtensions, currentMetadata?.userExtensions);
```
jacogr, line 347:
```suggestion
      registry.setSignedExtensions(request.payload.signedExtensions, currentMetadata?.userExtensions);
```

packages/extension-chains/src/index.ts
```diff
@@ -42,8 +42,10 @@ export function metadataExpand (definition: MetadataDef, isPartial = false): Cha
 
   if (metaCalls && !isPartial) {
     hasMetadata = true;
+    const metadata = new Metadata(registry, base64Decode(metaCalls));
```
jacogr, line 45:
```suggestion

    const metadata = new Metadata(registry, base64Decode(metaCalls));
```

jacogr (commented):
Small comments. But looks generally correct.

packages/extension-base/src/background/handlers/Extension.ts
KarishmaBothara: yes.. it did complain.. thanks for your suggestion.. made this change..

jacogr (approved):
Thank you.

PR #681, opened 2021-03-16
tr language support added

zinderud:
Hello, I added Turkish language support. I hope that will be useful


jacogr (commented):
Very useful indeed :)

We just need an entry in here as well - https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/util/getLanguageOptions.ts#L12-L27

(It is buried in the code, I have to find it myself)

jacogr (approved):
Thank you very much!

Issue #618, opened 2021-01-08
Polymath wallet interfering with extension?


anjaw3f:
A user reported that his polkadot.js extension only worked once he removed the Polymath Polymesh extension:

He had received the following error when trying to stake: 
_utility.batchAll
The source https://polkadot.js.org/apps/#/explorer has not been enabled yet_

I asked him to refresh the page to get the confirmation pop-up and allow the extension access, and this is what happened:

"Yes I was getting this pop-up and allowed the app to connect but it didn't work. That being said, another app was also giving me a pop-up. It's the Polymesh wallet (chrome extension) for Polymath token.
[..]
I went and deleted the polymesh wallet and it seems to have solved the issue. I was finaly able to stake my DOTs."

jacogr:
If he had the same address in both extensions, it could happen - i.e. the apps UI will pick the first one. The first error is since the page has not been enabled yet.

jacogr:
Closed in linked PR.

PR #685, opened 2021-03-24
✨ allows background handler to handle custom extention port name

hlminh2000:
Our custom wallet delegates handling of standard messages to polkadot-js library. The hard-coded reference to `PORT_EXTENSION` create a port name collision when users have multiple wallets installed. This change will allow us to provide the polkadot-js handler with our wallet's custom port name to avoid this collision.

Our target usecase: https://github.com/PolymathNetwork/polymesh-wallet/pull/126/files#diff-fdfafa0e47eeeae377557dd534aa80a02af8257fb509ab33f2d44b296e2b5bddL37

### DIFF
--- a/packages/extension-base/src/background/handlers/index.ts
+++ b/packages/extension-base/src/background/handlers/index.ts
@@ -14,8 +14,8 @@ const state = new State();
 const extension = new Extension(state);
 const tabs = new Tabs(state);
 
-export default function handler<TMessageType extends MessageTypes> ({ id, message, request }: TransportRequestMessage<TMessageType>, port: chrome.runtime.Port): void {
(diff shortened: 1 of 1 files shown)


jacogr:
Is this the same issue as in https://github.com/polkadot-js/extension/issues/618

hlminh2000:
@jacogr it seems definitely related. We have a similar issue on our Polymesh wallet where if the user has the DOT wallet installed, our wallet cannot perform its jobs. 
In our case, the issue is because of the port name collision, messages from the two wallets are indistinguishable on the page, so a race condition occurs for every message. Since the DOT wallet cannot process some of our wallet's messages, it would win the race with an error message and our page would fail with that error.

jacogr (approved):

PR #686, opened 2021-03-26
Prevent unauthorized apps from abusing pub(authorize.tab) request

remon-nashid:
An unauthorized app may send a large number of the "pub(authorize.tab)" request in a row to cause a denial of service, since each request launches a new popup. The app might send this request in a large volume that extension user never gets a chance to approve or reject those requests.


remon-nashid:
To produce the issue: You can paste this script in your browser's console while browsing a website that is NOT authorized yet. 

```
let n = 5;
var msg = {"id":"0.1","message":"","origin":"page","request":{"origin":"some app"}};
msg.message = "pub(authorize.tab)";
for(var i = 0; i < n; i++){
    window.postMessage(msg, "*");
}
```

jacogr (approved):
Thank you.

Issue #699, opened 2021-04-18
JSON encoding for account export broken when there's a # in the name


Tbaut:
Saw this bug in the wild: https://www.reddit.com/r/dot/comments/mtmpjc/cant_connect_my_wallet_extension_account_to_the/

How to replicate:
- create an account with a # in the name
- export the account as JSON
- see the broken JSON (at the end) e.g :
```
{"address":"5Cc..dw","encoded":"Dc8...xgS","encoding":{"content":["pkcs8","sr25519"],"type":["scrypt","xsalsa20-poly1305"],"version":"3"},"meta":{"genesisHash":"0x1611e1dbf0405379b861e2e27daa90f480b2e6d3682414a80835a52e8cb8a215","name":"rococo
```

jacogr:
Note sure if this belongs here or in the common or UI repos - will have to see where exactly.

jacogr:
Here is where it goes off the rails - https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/Popup/Export.tsx#L49-L51

Tbaut:
Ah so it's breaking because it's a href and # is a special char. I'm happy to make a PR for that. Should I use "file-saver" like in apps? I could research a bit to find some other way if we don't want to add this dependancy.

jacogr:
I read all the code for that module in 2 mins - it is tiny. Happy to have it in.

Tbaut:
Nice, I'll get to this most prob. this evening.

PR #701, opened 2021-04-21
Use file-saver to export accounts

Tbaut:
closes https://github.com/polkadot-js/extension/issues/699

It was very straightforward thanks to your pointers. I tested both export and exportAll on Chrome and FF.
~I simplified the test since it was complaining about the new return type of  `exportAccount`.  The mocked implementation was never used afaics.~
edit: Ah well, now I understand why it was there!


Tbaut:
That was too easy, the test throw:
```bash
console.error
    TypeError: i.createObjectURL is not a function
        at g (/home/runner/work/extension/extension/node_modules/file-saver/src/FileSaver.js:106:18)
        at /home/runner/work/extension/extension/packages/extension-ui/src/Popup/Export.tsx:49:11

      52 |         })
      53 |         .catch((error: Error) => {
    > 54 |           console.error(error);
         |                   ^
      55 |           setError(error.message);
      56 |           setIsBusy(false);
      57 |         });

      at packages/extension-ui/src/Popup/Export.tsx:54:19
```

which was reported a couple times in the Filesaver repo, e.g [here](https://github.com/eligrey/FileSaver.js/issues/611) but users went around using blob, which we're using. I'll see if there's something else.

jacogr (approved):
Looks sane, thank you!

PR #718, opened 2021-05-19
Support signing messages for external accounts

Tbaut:
I implemented the signing support in Stylo and submitted a PR for Signer.
It's working well with apps and Polkassembly.

https://user-images.githubusercontent.com/33178835/118889678-bce08700-b8fd-11eb-9e71-5fbf285d8082.mp4

### DIFF
--- a/packages/extension-ui/src/Popup/Signing/Qr.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Qr.tsx
@@ -48,11 +48,13 @@ function Qr ({ address, className, cmd, genesisHash, onSignature, payload }: Pro
   );
 
   if (!payloadU8a) {
-    return (<div className={className}>
-      <div className='qrContainer'>
-        Transaction command:{cmd} not supported.
+    return (
+      <div className={className}>
+        <div className='qrContainer'>
+          Transaction command:{cmd} not supported.
+        </div>
       </div>
-    </div>);
+    );
   }
 
   return (
--- a/packages/extension-ui/src/Popup/Signing/Qr.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Qr.tsx
@@ -2,39 +2,58 @@
 // SPDX-License-Identifier: Apache-2.0
 
 import type { ExtrinsicPayload } from '@polkadot/types/interfaces';
-import type { SignerPayloadJSON } from '@polkadot/types/types';
 
 import React, { useCallback, useMemo, useState } from 'react';
 import styled from 'styled-components';
 
 import { QrDisplayPayload, QrScanSignature } from '@polkadot/react-qr';
+import { hexToU8a } from '@polkadot/util';
 
 import { Button } from '../../components';
 import useTranslation from '../../hooks/useTranslation';
+import { CMD_MORTAL, CMD_SIGN_MESSAGE } from './Request';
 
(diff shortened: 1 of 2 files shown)


jacogr:
Sorry for the delay, will try to merge today - just one final look.

Tbaut:
No problem at all. I've added the genesis hash in the QR code here. We had a discussion with @Slesarew and although the genesis is not absolutely mandatory (after all the genesis shouldn't change anything in the signature), because the accounts are generally identified with a public key + genesis, on Signer/Stylo as well as the extension, it would make sense to keep it in message signing as well.

Slesarew:
> although the genesis is not absolutely mandatory 
Well, here, I must admit, it IS mandatory here after all. If we start signing arbitrary messages, that will be another uos call. The question is whether we always need another genhash in transactions, but it's out of scope here.

packages/extension-ui/src/Popup/Signing/Qr.tsx
```diff
@@ -2,50 +2,69 @@
...
     []
   );
 
+  if (!payloadU8a) {
+    return (<div className={className}>
+      <div className='qrContainer'>
+        Transaction command:{cmd} not supported.
+      </div>
+    </div>);
+  }
```
jacogr, line 56:
```suggestion
  if (!payloadU8a) {
    return (
      <div className={className}>
        <div className='qrContainer'>
          Transaction command:{cmd} not supported.
        </div>
      </div>
    );
  }
```

packages/extension-ui/src/Popup/Signing/Qr.tsx
```diff
@@ -2,50 +2,69 @@
...
 
   const _onShowQr = useCallback(
     () => setIsScanning(true),
     []
   );
 
+  if (!payloadU8a) {
+    return (<div className={className}>
+      <div className='qrContainer'>
+        Transaction command:{cmd} not supported.
```
jacogr, line 53: This should really be wrapped in `t('...')`, but I guess it is not absolutely critical ... if we do get here we probably have bigger issues...

packages/extension-ui/src/Popup/Signing/Request/index.tsx
```diff
@@ -123,20 +130,32 @@ export default function Request ({ account: { accountIndex, addressOffset, isExt
...
+              cmd={CMD_SIGN_MESSAGE}
+              genesisHash={account.genesisHash}
+              onSignature={_onSignature}
+              payload={data}
+            />
+          )
+          : <Bytes
+            bytes={data}
+            url={url}
+          />
```
jacogr, line 157:
```suggestion
          : (
            <Bytes
              bytes={data}
              url={url}
            />
          )
```

jacogr:
Some OCD-like comments, but seems sane.

packages/extension-ui/src/Popup/Signing/Qr.tsx
Tbaut: Thanks, totally, this slipped my mind. Will add it.

jacogr (approved):
As always, I really do appreciate your efforts. 

Exceptionally happy to see this in and supported, it has been a massive PITA with requests and support overhead.

PR #734, opened 2021-06-01
lint index.ts

joelamouche:
Separate PR for metamask extension, for formatting

### DIFF
--- a/packages/extension-dapp/src/index.ts
+++ b/packages/extension-dapp/src/index.ts
@@ -1,12 +1,23 @@
 // Copyright 2019-2021 @polkadot/extension-dapp authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


jacogr:
I went for the nuclear option and switched off prettier (via `.prettierignore`) on all polkadot-js repos. Sadly, the https://standardjs.com/ rules we based on and prettier just doesn't seem eye to eye. This is why we have multiple lint failures in this one. Some areas are configurable via the options, eg. quotes, but alas, not all - it is opinionated and sadly, so are the existing rules configured here over time based on the existing style.

Prettier certainly does do some nice stuff "in places", but beauty is in the eye of the beerholder - and in this case, there are a lot of areas which I just don't have the appetite for. For instance -

- the multi line wrapping for multiples from single-file imports
- the bracket spacing on functions (this is a clash with standard)
- it would re-print lines already split and below max length

I really should have switched on the ignore long time ago since I know a number of people do use it and it cases some to-and-fro due to irrelevant and wrong ("wrong" for the project and style)

At some point may re-evaluate, but as of now, there is no appetite for it. Hopefully the recent addition of the ignore file will help in the future.

joelamouche:
Ok good. So for my other PR, would running the linter solve my problems?

Issue #747, opened 2021-06-12
Support for signed ethereum messages


ntduan:
The signed messages are now wrapped in `<Bytes>`.So there is no way to sign messages for ethereum.

But actually, Ethereum messages have a prefix, "\x19Ethereum Signed Message:\n", so I think it can be handled specially, but when there is this prefix, it doesn't need to be wrapped in `<Bytes>`.

https://github.com/ethers-io/ethers.js/blob/master/packages/hash/src.ts/message.ts#L5

jacogr:
Indeed. It should be a simple addition, can just change this `isWrapped` to also detect the Ethereum form - https://github.com/polkadot-js/extension/blob/master/packages/extension-dapp/src/wrapBytes.ts#L11-L15

As always, PRs welcome.

PR #748, opened 2021-06-14
Don't wrap/unwrap Ethereum-style

jacogr:
Closes https://github.com/polkadot-js/extension/issues/747

### DIFF
--- a/packages/extension-dapp/src/wrapBytes.spec.ts
+++ b/packages/extension-dapp/src/wrapBytes.spec.ts
@@ -113,15 +113,6 @@ describe('unwrapBytes', (): void => {
     ).toEqual('');
   });
 
-  it('does not rewrap an Ethereum wrap', (): void => {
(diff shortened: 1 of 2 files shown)


packages/extension-dapp/src/wrapBytes.spec.ts
```diff
@@ -111,4 +112,33 @@ describe('unwrapBytes', (): void => {
...
+
+  it('does not rewrap an Ethereum wrap', (): void => {
+    expect(
+      u8aEq(
+        wrapBytes(TEST_ETH),
+        TEST_ETH
+      )
+    ).toBe(true);
+  });
+
```
jacogr, line 124:
```suggestion
```

jacogr:
Will make a path in an hour or two. (Which will include the latest util as well - no functional changes, but needs to tick along)

PR #767, opened 2021-07-03
Add configurable notifications

shawntabrizi:
This PR allows a user to configure how notifications from the extension are generated.

* none - No pop up is generated. The user can engage the extension directly.
* popup - A popup window is created, this is the current behavior.
* window - A new window is created.

Doesn't quite work yet, needs an update to the `ui-settings`: https://github.com/polkadot-js/ui/pull/498


packages/extension-base/src/background/handlers/State.ts
```diff
@@ -59,17 +60,30 @@ interface SignRequest extends Resolver<ResponseSigning> {
...
-const WINDOW_OPTS: chrome.windows.CreateData = {
-  // This is not allowed on FF, only on Chrome - disable completely
-  // focused: true,
+const NOTIFICATION_URL = chrome.extension.getURL('notification.html');
+
+const POPUP_WINDOW_OPTS: chrome.windows.CreateData = {
+  focused: true,
   height: 621,
   left: 150,
   top: 150,
```
shawntabrizi, line 69: note that these settings are ultimately the culprit of a really bad behavior. I dont see a good reason to specify these window sizes, since users can just manually control it themselves.

shawntabrizi: see: https://github.com/polkadot-js/extension/pull/768

jacogr (approved):
Thank you.

PR #768, opened 2021-07-03
Remove window sizes from window opts

shawntabrizi:
These window sizes have really poor behaviors (at least on MacOS using Firefox).

Without such specifications, users can control the window sizing on their own, and the browser will remember the size and location from previous.

Def recommend removing them.

### DIFF
--- a/packages/extension-base/src/background/handlers/State.ts
+++ b/packages/extension-base/src/background/handlers/State.ts
@@ -60,14 +60,9 @@ interface SignRequest extends Resolver<ResponseSigning> {
 let idCounter = 0;
 
 const WINDOW_OPTS: chrome.windows.CreateData = {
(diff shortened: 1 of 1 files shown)


packages/extension-base/src/background/handlers/State.ts
```diff
@@ -60,14 +60,9 @@ interface SignRequest extends Resolver<ResponseSigning> {
 let idCounter = 0;
 
 const WINDOW_OPTS: chrome.windows.CreateData = {
-  // This is not allowed on FF, only on Chrome - disable completely
-  // focused: true,
-  height: 621,
-  left: 150,
-  top: 150,
+  focused: true,
```
shawntabrizi, line 63: Focused is supported: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/create

Issue #758, opened 2021-06-21
Handle ethereum formated address in MetaData


joelamouche:
The way parchain address format are added into the browser extension or the mobile app doesn't seem to allow ethereum formated address, like we have in Moonbeam. 
Please let me know how to integrate this and I will help.

joelamouche:
@jacogr Is there a way to tell the extension that moonbeam address are supposed to be formated like ethereum? Should this be in Metadata?

joelamouche:
@jacogr let me know when you have time to look at this as it is becoming a time sensitive issue for us

jacogr:
Ok, so this is basically for chains added via the apps UI metadata. In that case the best would certainly be just to add another flag - I guess it is the isEthereum flag that should be transported. 

So would suggest making it a bit more generic, so pass something like an additional

chainType: “substrate” | “ethereum”

through in the metadata updates which uses the info from here https://github.com/polkadot-js/apps/blob/master/packages/page-settings/src/useChainInfo.ts#L15

joelamouche:
Here is a first PR to update the app : https://github.com/polkadot-js/apps/pull/5766
Let me know what part of extension I need to update to use ethereum formatted addresses

jacogr:
Will need to take a peek through, honestly not sure off-hand. Will take a browse through the code in the morning. 

I guess it would start with where the metadata is uploaded and start from there - with types, etc. The only issue here would be old/new, so don't want to break on _not_ having the field, but since we are mostly interested in the one, we can assume when not available it is the other.

jacogr:
https://github.com/polkadot-js/extension/blob/master/packages/extension-inject/src/types.ts#L52

It needs an additional `chainType?: 'substrate' | 'ethereum'` (it is marked optional since olde versions won't have the type, however that won't affect `def.chainType === 'ethereum'` checks anywhere, so all ok)

https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/components/Address.tsx#L101

The above handles the formatting, chain there should then have `chain.chainType?` which can be used in the component

joelamouche:
I made a first draft @jacogr https://github.com/polkadot-js/extension/pull/776
I just don't understand why the account doesnt appear in the app. Also the copy-paste doesnt work but that's less important

PR #776, opened 2021-07-07
Allow ethereum format for ethereum chains

joelamouche:
- Allows type injection from the current chain
- Allows creating ethereum formated addresses in the extension
- correctly injects ethereum address for signing

related to https://github.com/polkadot-js/extension/issues/758


joelamouche:
@jacogr here is a first draft. 
I was able to create an ethereum formated account but I don't understand why the account won't appear in the app

joelamouche:
@jacogr This was tested and works

crystalin:
Hi @jacogr, 
This is becoming critical for us. Is there something we can do to help getting this one merged ?

packages/extension-dapp/src/compat/singleSource.ts
```diff
@@ -73,15 +73,13 @@ function injectSingleSource (win: SingleWindow): void {
 // https://github.com/cennznet/singlesource-extension/blob/f7cb35b54e820bf46339f6b88ffede1b8e140de0/react-example/src/App.js#L19
 export default function initSingleSource (): Promise<boolean> {
   return new Promise((resolve): void => {
-    window.addEventListener('load', (): void => {
-      const win = window as Window & SingleWindow;
```
jacogr, line 77: I believe we need to drop this anyway, however this on-load used to be important here - before waiting for load, you would get inconsistencies - sometimes it would wok, sometimes it would not.

So this can go in (I really need to drop this, especially since we have the MM version available), but for reference the event listen here was there for a reason.

packages/extension-ui/src/Popup/CreateAccount/index.tsx
```diff
@@ -23,10 +23,18 @@ function CreateAccount ({ className }: Props): React.ReactElement {
...
-  const [genesis, setGenesis] = useState('');
+  const [genesisHash, setGenesis] = useState('');
+
+  function getChainName (genesisHash: string): string {
+    const filtered = options.filter(({ value }) => {
+      return genesisHash === value;
+    });
+
+    return filtered.length > 0 ? filtered[0].text : '';
+  }
```
jacogr, line 37: Just move this to file level, i.e. outside the React component. As tis stands the function would be re-created on each render.

packages/extension-ui/src/Popup/CreateAccount/index.tsx
```diff
@@ -81,9 +89,15 @@ function CreateAccount ({ className }: Props): React.ReactElement {
...
                   className={className}
                   label={t<string>('Network')}
-                  onChange={setGenesis}
+                  onChange={(newGenesisHash: string) => {
+                    if (getChainName(newGenesisHash) === 'Moonbase Alpha') { // TODO: use list
+                      setType('ethereum');
+                    }
+
+                    setGenesis(newGenesisHash);
+                  }}
```
jacogr, line 98:
```suggestion
                  onChange={(newGenesisHash: string) => {
                    if (getChainName(newGenesisHash) === 'Moonbase Alpha') { // TODO: use list
                      setType('ethereum');
                    }

                    setGenesis(newGenesisHash);
                  }}
```

We want this function defined inside the component with a `useCallback` definition. At this point it will be re-created each render, with the `useCallback` it will be crated once.

jacogr (commented):
Thank you.

Some small comments on moving 2 functions. Apart from that looks good.

packages/extension-dapp/src/compat/singleSource.ts
crystalin: @jacogr is there an efficient way to test reliably this part before merging ?

packages/extension-ui/src/Popup/CreateAccount/index.tsx
crystalin: @jacogr Thank you for reviewing.
I tried to quickly address your comment as Antoine is not available this week.
Let me know if there is more to do before you can merge it.

jacogr: Thank you.

packages/extension-dapp/src/compat/singleSource.ts
jacogr: Yes, you can test it against a locally built version of the extension. so 

- disable the one you have
- run yarn build
- add extension from path in Chrome

So if needed, what you will see is inconsistency - sometimes it may sow up, other times it may not show up. (All depending on load order, which is what this construct ensures.) It is a huge PITA, the extension itself also does this stuff on init and now that I mention it... it may just be enough and the above may indeed not be needed now...

crystalin: I don't feel to comfortable changing it, so I'm reverting it back. I'd prefer to have it in another PR

packages/extension-ui/src/Popup/CreateAccount/index.tsx
crystalin: I actually removed the function for simplicity. I think it will require a better design to support it if more chains are going to support Ethereum

packages/extension-ui/src/Popup/CreateAccount/index.tsx
```diff
@@ -40,20 +48,28 @@ function CreateAccount ({ className }: Props): React.ReactElement {
...
   const _onNextStep = useCallback(() => setStep((step) => step + 1), []);
   const _onPreviousStep = useCallback(() => setStep((step) => step - 1), []);
 
+  const _onChangeNetwork = useCallback((newGenesisHash: string) => {
+    if (getChainName(newGenesisHash) === 'Moonbase Alpha') { // TODO: use list
+      setType('ethereum');
+    }
+
+    setGenesis(newGenesisHash);
+  }, []);
```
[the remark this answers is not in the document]
crystalin: I fixed the lint

jacogr (approved):
Thank you. Lgtm, the chain matching is a tiny bit fragile here. This should however get solved (not here, but do-able now) since we do pass the `isEthereum` type now as part of the metadata updates.

jacogr:
Erk, seems to be a test failure, should be easy to address - https://github.com/polkadot-js/extension/pull/776/checks?check_run_id=3371419128#step:3:768

Change here - https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/Popup/CreateAccount/CreateAccount.test.tsx#L108

Just change `undefined` to `sr25519`

crystalin:
@jacogr I changed it, but I don't think it solves it.
I tried to look at what could be wrong, but it is too deep for me to grasp it quickly

jacogr:
No issues. Worst case I'll merge and fix whatever it is complaining about in a follow-up..

crystalin:
Thank you @jacogr 
(Ps: I sent you a message on Element)

crystalin:
Yeah !! 👍

packages/extension-dapp/src/compat/singleSource.ts
joelamouche: I removed the 'load' event because it was preventing the extension from working

Issue #830, opened 2021-09-17
Feat request: search bar


Tbaut:
Hello,
This is something I've been missing for a little while, I often find myself searching for an account visually and taking a while to find it. I'd benefit from a search bar, that would appear at the click of a "magnifying glass" button. I guess next to the + and cog wheel, and the search field would appear when the magnifying glass is clicked.

I'd be happy to take a look at it this week-end if you'd be ok with such a feature
![image](https://user-images.githubusercontent.com/33178835/133791673-3bdbbdad-eb6f-47d9-909e-1ce84d3fdb47.png)

jacogr:
Would be great. Thank you.

PR #832, opened 2021-09-18
Add account filtering feature

Tbaut:
closes https://github.com/polkadot-js/extension/issues/830

I did not implement the search by address, because I'd like to have your opinion first.
The account store has addresses with 42 prefix encoding, and searching by address would mean to 
- try to decode what the user types in the search field
- if this yields a public key, re-encode it using the 42 prefix
- filter based on this

I think this could really help, even novice users that have issues understanding the whole ss58 encoding, but this adds some overhead for most of the searches, which most probably won't be by address. What do you think?

Here's how it looks right now:

https://user-images.githubusercontent.com/33178835/133886994-ecc383c5-b453-4fa3-a8cc-240709eff4b4.mp4

### DIFF
--- a/packages/extension-ui/src/partials/Header.tsx
+++ b/packages/extension-ui/src/partials/Header.tsx
@@ -237,7 +237,6 @@ export default React.memo(styled(Header)(({ theme }: Props) => `
     }
   }
 
-
   .plusIcon, .cogIcon, .searchIcon {
     color: ${theme.iconNeutralColor};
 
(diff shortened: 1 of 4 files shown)


packages/extension-ui/src/partials/Header.tsx
```diff
@@ -154,29 +196,49 @@ export default React.memo(styled(Header)(({ theme }: Props) => `
...
+      width: 100%
+    }
+  
+    .popupToggle+.popupToggle {
+      margin-left: 8px;
+    }
   }
 
-  .plusIcon, .cogIcon {
+
```
Tbaut, line 240:
```suggestion
```

jacogr (approved):
Thank you!

PR #834, opened 2021-09-20
Add list of ethereum networks

joelamouche:
Unfortunately, I forgot to add Moonriver, so the extension compatibility only works with Moonbase Alpha...
@jacogr let me know if you can think of a better place to store this list (it is not the same as `import { ethereumChains } from '@polkadot/apps-config';`, which is runtime specname whereas in the extension it is the chain name)

### DIFF
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -297,7 +297,10 @@ export default class Extension {
   }
 
   private seedCreate ({ length = SEED_DEFAULT_LENGTH, type }: RequestSeedCreate): ResponseSeedCreate {
-    const seed = mnemonicGenerate(length);
+    let seed = mnemonicGenerate(length);
+    if (type==="ethereum"){
+      seed+=`/m/44'/60'/0'/0/0`
+    }
 
     return {
       address: keyring.createFromUri(seed, {}, type).address,
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -22,6 +22,9 @@ type CachedUnlocks = Record<string, number>;
 
 const SEED_DEFAULT_LENGTH = 12;
 const SEED_LENGTHS = [12, 15, 18, 21, 24];
+const ETH_DEFAULT_PATH = "m/44'/60'/0'/0/0";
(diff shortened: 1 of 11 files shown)


jacogr:
With the release cadence of the extension (it is purposefully really slow - (a) the approval process is not simple, (b) it requires a lot of testing and cannot be rushed), no sure a hard-coded list is going to work. 

Maybe the correct approach i to rather use the `isEthereum` that is passed as part of the metadata updates? That will always be current for chains that has been added at least once via metadata. (And we can warn if the chain being created for doesn't have any updated metadata as part of the creation process)

We can certainly still use a list as a fallback for known where there has been no metadata supplied, but would rather not rely on it as the be-all and end-all. Thoughts?

joelamouche:
> Thoughts?

Ok yes good idea. My plan is to use the chainType from metadata and a hardcoded list as fallback

joelamouche:
I want to add this feature: https://github.com/polkadot-js/extension/issues/837

joelamouche:
@jacogr 
With the ethereum (bip44) derivation path, the button is too low to be seen
Screenshot 2021-09-23 at 15 40 13
How can I change the window size or make it scrollable?
Also I want to hide the derivation path components in an Expander, as it is done in the polkadot-js app, but that component is only in `@polkadot/react-components`.
Should I :
- import from `@polkadot/react-components`
- add it to extension components
- other

joelamouche:
@jacogr pls review when you get the opportunity

joelamouche:
@jacogr please take the time to respond to my comment https://github.com/polkadot-js/extension/pull/834#issuecomment-925828782 when you have the time

joelamouche:
@jacogr please take the time to respond to my comment #834 (comment) when you have the time

crystalin:
@jacogr ,as the launch of parachain on Polkadot is coming soon, it'd be great if we can offer support of polkadotjs extension.

jacogr:
Replaced by https://github.com/polkadot-js/extension/pull/882

(doesn't contain hard-coded lists, rather relies on the chain type in the metadata. Tested on Moonriver for both import and new creation)

joelamouche:
@jacogr Thanks! What about the CreateEthDerivationPath component? Using ethereum stye derivation with derivation path is important because people are creating addresses with the wrong derivation

jacogr:
100%, that should be added.

jacogr:
That would make 100% sense.

PR #845, opened 2021-10-01
Added urdu to language options

itsonal:


jacogr (commented):
There is no Urdu translation - https://github.com/polkadot-js/extension/tree/master/packages/extension/public/locales

jacogr (changes requested):
This should be done alongside adding a translation.json for for the language. The option dictates which language file gets used.

Issue #858, opened 2021-10-19
Make base-extension more flexible 


AndreiEres:
I'm developing an extension based on that extension. And I see that datas from my and original extension are mixing. To prevent it I had to copy-paste code from base-extension and rewrite some identifiers like port names. That's not a good Idea of course. 

What do you think about adding some prefixes from env-vars to that sensible names? Or maybe we can use another way to use base extension in other extensions without monkey-patching?

AndreiEres:
I added an extension prefix to separate another extension's states and ports. What do you think?

PR #891, opened 2021-11-17
Add extension prefix

AndreiEres:
Fixes https://github.com/polkadot-js/extension/issues/858

### DIFF
--- a/packages/extension-base/src/background/types.ts
+++ b/packages/extension-base/src/background/types.ts
@@ -141,7 +141,7 @@ export type MessageTypesWithNullRequest = NullKeys<RequestTypes>
 export interface TransportRequestMessage<TMessageType extends MessageTypes> {
   id: string;
   message: TMessageType;
-  origin: 'page' | 'extension';
+  origin: string;
   request: RequestTypes[TMessageType];
 }
 
--- a/packages/extension-base/src/defaults.ts
+++ b/packages/extension-base/src/defaults.ts
@@ -2,9 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 const ALLOWED_PATH = ['/', '/account/import-ledger', '/account/restore-json'] as const;
-const PORT_CONTENT = 'content';
 const PHISHING_PAGE_REDIRECT = '/phishing-page-detected';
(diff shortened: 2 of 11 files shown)


jacogr:
Sorry have not forgotten about this, will most probably get to reviewing over the weekend.

mrcali213 (approved):

mrcali213 (approved):

mrcali213 (approved):

mrcali213 (approved):

packages/extension-base/src/background/types.ts
```diff
@@ -141,7 +141,7 @@ export type MessageTypesWithNullRequest = NullKeys<RequestTypes>
 export interface TransportRequestMessage<TMessageType extends MessageTypes> {
   id: string;
   message: TMessageType;
-  origin: 'page' | 'extension';
+  origin: string;
```
jacogr, line 144:
```suggestion
  origin: `${string}${'page' | 'extension'}`;
```

... but also no need for magic, so 100% fine and correct as-is

packages/extension-base/src/defaults.ts
```diff
@@ -2,9 +2,12 @@
 // SPDX-License-Identifier: Apache-2.0
 
 const ALLOWED_PATH = ['/', '/account/import-ledger', '/account/restore-json'] as const;
-const PORT_CONTENT = 'content';
 const PHISHING_PAGE_REDIRECT = '/phishing-page-detected';
-const PORT_EXTENSION = 'extension';
+const EXTENSION_PREFIX = process.env.EXTENSION_PREFIX as string || '';
```
jacogr, line 6:
```suggestion
const EXTENSION_PREFIX = typeof process === 'object' && typeof process.env === 'object' ? (process.env.EXTENSION_PREFIX as string || '') : '';
```

Small suggestion, had some weird stuff with environments breaking in the past - however, this also should be optional, at this level if/when it does break the users should know how to fix it. (Since the target here is extension devs)

jacogr (approved):
Brilliant! Thanks an absolute million, it looks spot-on.

PR #1005, opened 2022-02-18
fix(authorize): long urls pushes the cta down

juliuscarvajal:
This PR fixes the issue where a long url breaks the Authorize popup window.
Screen Shot 2022-02-18 at 4 52 05 pm
Screen Shot 2022-02-18 at 4 52 42 pm


jacogr (approved):
Thank you. These long URLs we probably need to trim to some length, max 64 chars with ellipsis. (I think all the details, especially with params, is probably not that important as long as we have the domain + path)

juliuscarvajal:
Thanks for the quick response @jacogr. Is it possible to have a release to chrome store for this?
The long urls are not allowing the user to proceed their workflow onboarding on Collabland via Polkadotjs.

PR #1012, opened 2022-03-02
Add option for selecting crypto type on import

nathanwhit:
This PR adds a dropdown to the "advanced" section of the "Import account" dialog that allows selecting the keypair's crypto type. It was already possible to import other keypair types by restoring from a JSON file, now it is possible to import other keypair types by seed.


jacogr:
I'd rather not add this since it just complicates the world. The extension has always been for the Polkadot-recommended sr25519 only (except for ethereum-chains), specifically to ensure that all the noise found elsewhere around "which type and where" doesn't exist here. There is not meant to be "advanced" options in the extension. It needs to be simplified more, not expanded.

jacogr:
Closing. Due to the above, we really want to keep this on non-user-selectable with sane defaults. (As-is it is already too complicated)

PR #1049, opened 2022-04-22
Add Bengali translation

itsonal:
Bengali is the fifth highest spoken language in the world. Its the national language of Bangladesh and spoken widely in India too. 
https://en.wikipedia.org/wiki/Bengali_language


jacogr (approved):
Thank you. As with the apps UI, making it active is _also_ slightly counter-intuitive... :) In this case, you need to add an entry here - https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/util/getLanguageOptions.ts

PR #1055, opened 2022-04-27
Present the correct signing data in the signing page

hamidra:
Currently the signing data content that is presented in the extension to be signed by the user removes the white space and new lines,
Screen Shot 2022-04-27 at 2 03 59 PM

This is not very user friendly and also might have security implications since the users are not presented with the actual data that are signing.

this changes make the signing data a scrollable multiline pre box to show the exact data in the signing box.
Screen Shot 2022-04-27 at 2 08 44 PM


packages/extension-ui/src/Popup/Signing/Bytes.tsx
```diff
@@ -53,11 +54,24 @@ export default styled(Bytes)`
     text-overflow: ellipsis;
     vertical-align: middle;
     width: 100%;
+    padding: 5px;
```
jacogr, line 57: `0.125rem`? Just prefer to stick to one unit, can obviously adjust slightly.

packages/extension-ui/src/Popup/Signing/Bytes.tsx
```diff
@@ -53,11 +54,24 @@ export default styled(Bytes)`
...
 
-    pre {
-      font-family: inherit;
-      font-size: 0.75rem;
-      margin: 0;
+    &.pre {
+      padding: 0px;
+      
+      div {
+        padding: 5px;
```
jacogr, line 63: As above, maybe `0.125rem`?

packages/extension-ui/src/components/Address.tsx
```diff
@@ -306,7 +306,7 @@ function Address ({ actions, address, children, className, genesisHash, isExtern
 }
 
 export default styled(Address)(({ theme }: ThemeProps) => `
-  background: ${theme.accountBackground};
+  background: ${theme.boxBackground};
```
jacogr, line 309: Since it is only used in accounts, not sure the rename is needed? (It did refer to location)

jacogr (commented):
Needs CI fixes. (`yarn lint` locally could auto-fix most/all, have not looked at the specifics)

packages/extension-ui/src/Popup/Signing/Bytes.tsx
```diff
@@ -53,11 +55,24 @@ export default styled(Bytes)`
...
+        padding: 0.15rem;
+        font-family: inherit;
+        font-size: 0.75rem;
+        margin: 0;
+        white-space: pre;
+        overflow: auto;
+        max-height: calc(100vh - 480px);
+        min-height: ${theme.boxLineHeight};
+        border: 1px solid ${theme.boxBorderColor};
+        background: ${theme.boxBackground};
```
hamidra, line 73: used the boxBackground theme here.

packages/extension-ui/src/components/Address.tsx
hamidra: I also used this for the signing data box background. Do you prefer to have different theme variables for account and box or just having a generic box background to be used for all type of these boxes makes sense?!

packages/extension-ui/src/Popup/Signing/Bytes.tsx
jacogr: Ok, in that case, makes sense - not singular use.

jacogr (approved):
Thank you.

Issue #1058, opened 2022-05-06
Settings menu and + menu do not appear on click, on master


Tbaut:
On master build, the settings menu and + menu do not appear when clicking on them. This looks like some styling issues. I'm not sure what update could have caused this. I went back to the commit 415184ac "Add authorized URL removal  (#1029)" and it's working there.

Tested on Chrome and FF.

jacogr:
Weird :( I should have picked this up since I generally run master, but behind a bit :(

Tbaut:
I'm working on the auth flow for now but I will check it out in the next days

jacogr:
I will check this one out, no sweat.

jacogr:
The `useOutsideClick` seem to now screw things up. When I comment it out, it shows. So it would seem the event fires on both those and just cancels the show.

I think this is indeed React 18 related, it batches the updates.

PR #1059, opened 2022-05-07
Adjust useOutsideClick with multi-references

jacogr:
Closes https://github.com/polkadot-js/extension/issues/1058

Because of the state batching, you need to check both the icon and menu references to ensure that we don't untoggle.

cc @Tbaut - a check would be worth-while, it does seem to fix it for me.


Tbaut (approved):
:100: working perfectly on Brave and FF for me now, thanks a lot

PR #1072, opened 2022-05-27
can not read the length of an undefined

Nick-1979:
In some cases (when computer is busy or have many accounts), throw the error :' can not read the length of an undefined'
hence changed hierarchy.length ===> hierarchy?.length

### DIFF
--- a/packages/extension-ui/src/Popup/Accounts/index.tsx
+++ b/packages/extension-ui/src/Popup/Accounts/index.tsx
@@ -43,7 +43,7 @@ function Accounts ({ className }: Props): React.ReactElement {
 
   return (
     <>
-      {(hierarchy.length === 0)
+      {(hierarchy?.length === 0)
         ? <AddAccount />
         : (
           <>


jacogr:
This doesn't quite make sense - the check here does, however what doesn't make sense -

```
const AccountContext = React.createContext<AccountsContext>({ accounts: [], hierarchy: [], master: undefined });
```

So there _must_ always be a hierarchy. Would rather try and track down why it becomes `undefined` (i.e. the root cause) and get a fix for that.

packages/extension-ui/src/Popup/Accounts/index.tsx
```diff
@@ -43,7 +43,7 @@ function Accounts ({ className }: Props): React.ReactElement {
 
   return (
     <>
-      {(hierarchy.length === 0)
+      {(hierarchy?.length === 0)
```
jacogr, line 46:
```suggestion
      {!hierarchy?.length
```

The original doesn't quite do the trick. since `undefined !== 0`

Actually, what this most-probably needs to do is something like -

```js
if (! hierarchy) {
  return null;
}
```

Since we most probably don't want anything to show when we don't know, i.e. things are not in-place.

PR #1076, opened 2022-05-31
adding option to choose wallet to connect

pedroapfilho:
This PR enables the possibility of enabling only a specific wallet if needed.

I'm creating it to be used here: https://github.com/Premiurly/polkassembly/pull/132

### DIFF
--- a/packages/extension-dapp/src/bundle.ts
+++ b/packages/extension-dapp/src/bundle.ts
@@ -30,15 +30,13 @@ function throwError (method: string): never {
 
 // internal helper to map from Array<InjectedAccount> -> Array<InjectedAccountWithMeta>
 function mapAccounts (source: string, list: InjectedAccount[], ss58Format?: number): InjectedAccountWithMeta[] {
-  return list.map(({ address, genesisHash, name, type }): InjectedAccountWithMeta => {
-    const encodedAddress = address.length === 42 ? address : encodeAddress(decodeAddress(address), ss58Format);
-
-    return ({
-      address: encodedAddress,
-      meta: { genesisHash, name, source },
-      type
-    });
-  });
+  return list.map(({ address, genesisHash, name, type }): InjectedAccountWithMeta => ({
+    address: address.length === 42
+      ? address
+      : encodeAddress(decodeAddress(address), ss58Format),
+    meta: { genesisHash, name, source },
+    type
+  }));
 }
 
 // have we found a properly constructed window.injectedWeb3
@@ -49,9 +47,9 @@ let web3EnablePromise: Promise<InjectedExtension[]> | null = null;
 
 export { isWeb3Injected, web3EnablePromise };
 
-function getWindowExtension (originName: string, extensionName: string): Promise<[InjectedExtensionInfo, Injected | void][]> {
+function getWindowExtension (originName: string, extensionName?: string): Promise<[InjectedExtensionInfo, Injected | void][]> {
   return Promise.all(
-    Object.entries(win.injectedWeb3).filter(() => win.injectedWeb3[extensionName]).map(
+    Object.entries(win.injectedWeb3).filter(() => extensionName ? win.injectedWeb3[extensionName] : true).map(
       ([name, { enable, version }]): Promise<[InjectedExtensionInfo, Injected | void]> =>
         Promise.all([
           Promise.resolve({ name, version }),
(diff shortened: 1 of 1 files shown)


fend25:
Hi! Very usefull and needed addition. Actually I visited PRs page to see whether there is some PR like this one and start my own if I don't find one. Thank you, Pedro.

@jacogr hope you'll find couple of minutes to review this tiny, but very awaited PR.

Have a nice day!

packages/extension-dapp/src/bundle.ts
```diff
@@ -47,9 +47,9 @@ let web3EnablePromise: Promise<InjectedExtension[]> | null = null;
 
 export { isWeb3Injected, web3EnablePromise };
 
-function getWindowExtensions (originName: string): Promise<[InjectedExtensionInfo, Injected | void][]> {
+function getWindowExtension (originName: string, extensionName?: string): Promise<[InjectedExtensionInfo, Injected | void][]> {
   return Promise.all(
-    Object.entries(win.injectedWeb3).map(
+    Object.entries(win.injectedWeb3).filter(() => extensionName ? win.injectedWeb3[extensionName] : true).map(
```
jacogr, line 52: This is not quite correct. You don't actually filter the entries based on the name retrieved.

```js
Object.entries(win.injectedWeb3).filter(([name]) => !extensionName || name === extensionName)
```

packages/extension-dapp/src/bundle.ts
```diff
@@ -61,8 +61,8 @@ function getWindowExtensions (originName: string): Promise<[InjectedExtensionInf
   );
 }
 
-// enables all the providers found on the injected window interface
-export function web3Enable (originName: string, compatInits: (() => Promise<boolean>)[] = []): Promise<InjectedExtension[]> {
+// enables all the providers found on the injected window interface, or just the named one
+export function web3Enable (originName: string, extensionName?: string, compatInits: (() => Promise<boolean>)[] = []): Promise<InjectedExtension[]> {
```
jacogr, line 65:
```suggestion
export function web3Enable (originName: string, compatInits: (() => Promise<boolean>)[] = [], extensionName?: string): Promise<InjectedExtension[]> {
```

Should be added to the back since moving it around would break it for users that already use the compat (and don't use type checking)

jacogr (commented):
Some (small) changes required.

packages/extension-dapp/src/bundle.ts
pedroapfilho: Just to confirm, that I can just pass an empty array and keep the same behaviour as not having it

jacogr: 100%. Empty array means there is nothing to apply.

pedroapfilho:
@jacogr all fixed!

JurbinM (approved):

pedroapfilho:
@jacogr could you check this one again?

Rocket19-1:
Hi you guys are the writers and builders I'm on the outside looking into this multiple wallets with multiple functions creates more work costing data people time so logic tells me that humanity likes things simple with ahassle free time saving components so a wallet that can perform all of these  different functions is a move in the rite direction I feel if this could be developed and phase out multiple wallets people will naturally progress towards it it may be complicated and complex but once this is in place I think benifits will be notified

pedroapfilho:
Hey @jacogr is it possible for you to check this out again?

jacogr:
As mentioned above - this really flies in the face of what the interface is meant to be. 

It is designed so that you don't need to provide the user with a dialog to select a specific provider. (This in my view is a hold-over from the old days in other environments where each had their own interface) 

It is meant to happily pull in accounts from multiple providers and just allow you to use them.

0xKheops:
I have to agree with Jaco, sorry Pedro :heart: 

Users should decide which wallet he wants to use, not the dapp.

Having per-wallet button forces the dapp to have one button per existing wallet. it's sustainable now as there are very few polkadot compatible wallets, but it won't be in a year or two.
Whether it's on purpose or because the dapp developer doesn't know about the existence of a given wallet, some users will be left out.

Enabling a specific wallet is already achievable by doing `window.injectedWeb3["the-wallet-id"].enable("dapp-name")`. It may not be as intuitive but still is easy enough.
There's even a popular dapp right now that only allows 1 single wallet to connect (allowing to vote on which is the best wallet :joy:),  this is not something that should be encouraged by the mainstream API.

jacogr:
`window.injectedWeb3["the-wallet-id"].enable("dapp-name")` will actually stop being available in a future version of the interfaces. Let me explain...

The `window.injectedWeb3` object does "leak" information about what is available to the specific user. When everybody plays nice, not an issue, however it does show all sites what is there without the user even having taken action to enable/disable. So this needs to be more opaque until `enable` (or something like enable) indicates stuff is approved by the user and the underlying info is returned.

So in a future version the exposed interface will change where the keys are being hashed (and being different on each load) and the `enable` call will actually return the extension info. Obviously the old way will still be there for extensions that does not support it yet, however certainly for the polkadot-js extension we will move in that direction as soon as available as an option on extension-dapp and extension-inject. (This would be around January-ish, just need to get the Manifest 3 conversion done this year.)

EDIT:

pedroapfilho:
That's very cool! Thanks for all the explanations! I'm happy to see that we are moving away from it.

Issue #1061, opened 2022-05-10
Sign transactions on moon* chains


ekenigs:
I have tried to figure out what needs to be done so users can sign transactions on Moonbase, Moonriver, and Moonbeam chains from the Extension on Polkadot.js Apps. I'm a bit stuck with my investigation and need some help.

Currently, there is an error when I'm trying to send funds to another account:
<img width="835" alt="Screenshot 2022-05-10 at 10 59 39" src="https://user-images.githubusercontent.com/15612217/167693082-8fa599bd-bbb1-43e3-973d-dbfa757865ad.png">

The difference in signing transactions from Apps or Extension is that a prefix is added when Extension is signing.

https://github.com/polkadot-js/common/blob/master/packages/keyring/src/pair/index.ts#L190

```typescript
const TYPE_PREFIX = {
  ecdsa: new Uint8Array([2]),
  ed25519: new Uint8Array([0]),
  ethereum: new Uint8Array([2]), // <- this
  sr25519: new Uint8Array([1]),
};
```
```typescript
    sign: (message: HexString | string | Uint8Array, options: SignOptions = {}): Uint8Array => {
      assert(!isLocked(secretKey), 'Cannot sign with a locked key pair');

      return u8aConcat(
        options.withType
          ? TYPE_PREFIX[type]
          : SIG_TYPE_NONE,
        TYPE_SIGNATURE[type](u8aToU8a(message), { publicKey, secretKey })
      );
    },
```

The flag `options.withType` is coming from:

https://github.com/polkadot-js/api/blob/master/packages/types/src/extrinsic/v4/ExtrinsicPayload.ts#L36

```typescript
  constructor (registry: Registry, value?: ExtrinsicPayloadValue | Uint8Array | HexString) {
    super(registry, objectSpread(
      { method: 'Bytes' },
      registry.getSignedExtensionTypes(),
      registry.getSignedExtensionExtra()
    ), value);

    // Do detection for the type of extrinsic, in the case of MultiSignature
    // this is an enum, in the case of AnySignature, this is a Hash only
    // (which may be 64 or 65 bytes)
    this.#signOptions = {
      withType: registry.createTypeUnsafe('ExtrinsicSignature', []) instanceof Enum
    };
  }
```

What I don't understand is why when Apps is signing then `ExtrinsicSignature` is not an instance of `Enum` and when from Extension it is.

I have tried to work around it by changing the condition to:

```typescript
      return u8aConcat(
[...]

Issue #907, opened 2021-12-06
Unable to Sign Transactions from Injected Ethereum-Style Accounts


themacexpert:
I am unable to successfully sign any transactions in Polkadot Js Apps using an injected Ethereum-style account from the extension. 

Steps to Re-Create:
- Create a new Account in the Polkadot Js extension that uses an Ethereum-style address. Moonbase Alpha or Moonriver for Example.
- To avoid using real funds, you can get Moonbase Alpha tokens in the #moonbase-faucet channel of our discord here: https://discord.gg/moonbeam. You can request funds using the format `!faucet send <your address>`
- Go to Polkadot Js Apps and update the Metadata for the Extension.
- Try to send funds from the injected account to any address including itself. You'll get the below error.
![Screen Shot 2021-12-06 at 2 51 46 PM](https://user-images.githubusercontent.com/5834262/144935583-f861b355-4fc1-4349-a3d0-7200be458244.png)

Looking at the console I see the following messages:
![Screen Shot 2021-12-06 at 2 54 46 PM](https://user-images.githubusercontent.com/5834262/144935894-af8e3c2f-0480-49e9-b31c-efd7ccdbf9d1.png)

This issue has been replicated by several individuals with the same results.

jacogr:
Would need to look at the Moonbeam transaction signature type, e.g. `MultiSignature` or `AnySignature`, my guess is that is where it would break down. (But would then apply to all, not really "only extension accounts")

So generally the above is either -

- signing payload incorrect
- signature is in the wrong format (e.g. MultiSignature is the signature + a single leading byte)
- ... other?

There are tricky. If it only happens on extension accounts, it would mean that the extension does something weird. (Well, could also be in the apps UI signer where it uses the signature from the extension)

TL;DR Only ramblings above... this is not something that is clear in what may be the issue.

joelamouche:
@jacogr Any news on this?

>signing payload incorrect

I added a test to the extension repo to make sure that is not the problem

>signature is in the wrong format

How do we test that? Unfortunately this is hidden in the `.sign` function

joelamouche:
@jacogr Any news on this? This is still a problem

joelamouche:
I havent been able to find the solution to this but it is pretty important to allow new users to join in. I would love to get on a pair programming session

hoale58-vl:
Hello guys, 
[...]

PR #1090, opened 2022-06-15
Updating registry with chain metadata

ekenigs:
This is a fix for issues https://github.com/polkadot-js/extension/issues/907 and https://github.com/polkadot-js/extension/issues/1061.

With updated metadata registry is providing a correct type for `ExtrinsicSignature` and the extension is signing transactions correctly.


packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -232,6 +232,10 @@ export default class Extension {
     const { request, resolve } = queued;
 
     this.#state.saveMetadata(request);
+    request.metaCalls &&
```
jacogr, line 235: This should not happen on save. It should kick in for the instance where the signing takes place. 

The issue you have here is that each update injects their own and they are different.

ekenigs: Thanks, @jacogr  for your fast feedback. Moved to the signing method.

ekenigs:
@jacogr hey, can you please check this?

crystalin:
@jacogr gentle ping :)

JurbinM (approved):

ekenigs:
@JurbinM hey, thanks for reviewing, can you please run checks and merge PR?

PR #1068, opened 2022-05-17
Account selection when authorizing a website

Tbaut:
closes https://github.com/polkadot-js/extension/issues/1037

What is there:
- account selection
- un/toggle all accounts
- show a dedicated message when no account was found:  [screenshot](https://user-images.githubusercontent.com/33178835/169050614-f1ce56bd-076e-4ace-b3fd-7e02b1d0c6d5.png)
- refactor the website management screen to allow changing the connected accounts

It looks like this:

https://user-images.githubusercontent.com/33178835/173256232-a713a78f-deaa-4678-803f-b1b0fbd190f2.mp4

https://user-images.githubusercontent.com/33178835/169295924-94d29491-fbe8-4e8b-ae5f-df1e9cac873d.mp4

TODO:

- [x] Show helper text & button when the extension doesn't have any account yet.
- [x] Fix the FIXME
- [x] Fix/Add tests
- [x] Add account management per website from the "Manage Website Access" screen.
- [x] Testing on FF and with less accounts
- [x] When an account is forgotten, we need to go through the list of authorized website and remove the said account.

PR #1110, opened 2022-07-14
Handle migration from old all-accounts

jacogr:
Follow-up for https://github.com/polkadot-js/extension/pull/1068

Tested by going to 0.44.1, reject/approve and then running latest

### DIFF
--- a/packages/extension-base/src/background/handlers/State.ts
+++ b/packages/extension-base/src/background/handlers/State.ts
@@ -35,6 +35,8 @@ export type AuthorizedAccountsDiff = [url: string, authorizedAccounts: AuthUrlIn
 export interface AuthUrlInfo {
   count: number;
   id: string;
+  // this is from pre-0.44.1
+  isAllowed?: boolean;
   origin: string;
   url: string;
   authorizedAccounts: string[];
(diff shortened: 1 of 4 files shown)


Tbaut:
Interesting, I did not handle the migration as we discussed in the original issue so that users would have to go through the new process and decide whether or not all accounts should be shared.

jacogr:
The issue is, once I rolled back to the branch, it was confusing - nothing popped up, but 0 accounts were injected. So I do recall the re-auth, but assumed it would be a popup that asks for it. It never happened. This seemed to be the simplest way around that...

packages/extension-base/src/background/handlers/State.ts
```diff
@@ -368,19 +367,20 @@ export default class State {
...
-    const isDuplicate = Object.values(this.#authRequests)
+    const isDuplicate = Object
+      .values(this.#authRequests)
       .some((request) => request.idStr === idStr);
 
     assert(!isDuplicate, `The source ${url} has a pending authorization request`);
 
     if (this.#authUrls[idStr]) {
       // this url was seen in the past
-      assert(this.#authUrls[idStr].authorizedAccounts, `The source ${url} is not allowed to interact with this extension`);
```
Tbaut, line 378: good point, I must have been missing this in my tests, I expected this part to work for the migration since any old website would have no `authorizedAccounts`, a check for `=== undefined` may have worked as well. But your solution is even more elegant anyway.

Issue #1118, opened 2022-07-26
Connect + account select -> "Ask again later" button is hidden


jacogr:
... at least for me on Chrome. Needs more spacing in here.

Tbaut:
I'll take a look

PR #1119, opened 2022-07-27
Fix ask again invisible in popup

Tbaut:
closes https://github.com/polkadot-js/extension/issues/1118

While I couldn't reproduce it in the initial popup, I could see the ask again hidden in the popup that appears when clicking on the extension icon. I bet that's what you meant, I missed this originally, sorry about that. I did some small adjustment to have it visible.

image

### DIFF
--- a/packages/extension-ui/src/Popup/Authorize/Request.tsx
+++ b/packages/extension-ui/src/Popup/Authorize/Request.tsx
@@ -84,15 +84,16 @@ function Request ({ authId, className, isFirst, request: { origin }, url }: Prop
 export default styled(Request)`
   .acceptButton {
     width: 90%;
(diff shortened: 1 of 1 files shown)


jacogr:
In the extension click, it is half-hidden. When it is a window on a new site, it is completely hidden for me. (This is on Chrome)

Will check this out, thank you.

Tbaut:
I remember the pain between FF and Chrome with the window popup compared to the extension click, but I didn't remember things being different based on the OS :see_no_evil: 
It was half hidden for me only on the extension click, for chrome, FF and Brave, so at least this was consistent. This removes small margin here and there to make it fully visible for me, hope that does the trick for you too.

jacogr (approved):
Thank you.

PR #1126, opened 2022-08-24
Add isEnabled call to get extension authorization status

SaltyCucumber:
Currently there is no way of checking if extension is already authorized without calling a popup, so most dapp devs are holding status in localstorage or similar, which is a workaround and adds unnecessary code.

This PR solves that by adding isEnabled call to extension-inject, which can be used to get the extension status.

UseCase:
As a dapp dev I want to support multiple extensions, show their status connected/disconnected without calling .enable for all extensions at the same time and creating multiple popups, which is not a great UX.
I also don't want to be creating a custom solution to this problem and maintain it.

### DIFF
--- a/packages/extension-compat-metamask/src/bundle.ts
+++ b/packages/extension-compat-metamask/src/bundle.ts
@@ -85,7 +85,7 @@ function injectMetaMaskWeb3 (win: Web3Window): void {
       };
     },
     isEnabled: async (): Promise<boolean> => {
-      return new Promise((resolve) => resolve(true));
+      return Promise.resolve(true);
     }, // TODO: stub for now
     version: '0' // TODO: win.ethereum.version
   };
--- a/packages/extension-compat-metamask/src/bundle.ts
+++ b/packages/extension-compat-metamask/src/bundle.ts
@@ -84,6 +84,9 @@ function injectMetaMaskWeb3 (win: Web3Window): void {
         }
       };
     },
+    isEnabled: async (): Promise<boolean> => {
+      return new Promise((resolve) => resolve(true));
+    }, // TODO: stub for now
     version: '0' // TODO: win.ethereum.version
   };
 }
--- a/README.md
+++ b/README.md
@@ -120,6 +120,7 @@ window.injectedWeb3 = {
 
     // this call returns extension status that tells if extension is already authorized
     isEnabled (): Promise<boolean>,
+
     // semver for the package
     version: '0.1.0'
   }
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ As it stands, it does one thing: it _only_ manages accounts and allows the signi
 ![interface screenshots](docs/extension-overview.png)
 
 ## Documentation and examples
-Find out more about how to use the extension as a Dapp developper, cookbook, as well as answers to most frequent questions in the [Polkadot-js extension documentation](https://polkadot.js.org/docs/extension/)
+Find out more about how to use the extension as a Dapp developer, cookbook, as well as answers to most frequent questions in the [Polkadot-js extension documentation](https://polkadot.js.org/docs/extension/)
 
 ## Development version
 
--- a/README.md
+++ b/README.md
@@ -113,13 +113,15 @@ window.injectedWeb3 = {
   // this is the name for this extension, there could be multiples injected,
   // each with their own keys, here `polkadot-js` is for this extension
   'polkadot-js': {
-    // semver for the package
-    version: '0.1.0',
-
     // this is called to enable the injection, and returns an injected
     // object containing the accounts, signer and provider interfaces
     // (or it will reject if not authorized)
-    enable (originName: string): Promise<Injected>
+    enable (originName: string): Promise<Injected>,
+
+    // this call returns extension status that tells if extension is already authorized
+    isEnabled (): Promise<boolean>,
+    // semver for the package
+    version: '0.1.0'
   }
 }
 ```
--- a/packages/extension-base/src/background/handlers/Tabs.ts
+++ b/packages/extension-base/src/background/handlers/Tabs.ts
@@ -215,11 +215,14 @@ export default class Tabs {
       return this.redirectIfPhishing(url);
     }
 
-    if (type !== 'pub(authorize.tab)') {
+    if (type !== 'pub(authorize.tab)' && type !== 'pub(authorize.isEnabled)') {
       this.#state.ensureUrlAuthorized(url);
     }
 
     switch (type) {
+      case 'pub(authorize.isEnabled)':
+        return !!this.#state.authUrls[this.#state.stripUrl(url)];
+
       case 'pub(authorize.tab)':
         return this.authorize(url, request as RequestAuthorizeTab);
 
--- a/packages/extension-base/src/background/types.ts
+++ b/packages/extension-base/src/background/types.ts
@@ -119,6 +119,7 @@ export interface RequestSignatures {
   'pub(accounts.list)': [RequestAccountList, InjectedAccount[]];
   'pub(accounts.subscribe)': [RequestAccountSubscribe, string, InjectedAccount[]];
   'pub(accounts.unsubscribe)': [RequestAccountUnsubscribe, boolean];
+  'pub(authorize.isEnabled)': [null, Promise<boolean>];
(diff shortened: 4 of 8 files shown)


jacogr:
I don't get it - when enabled, there are no extension popups when calling `web3Enable`.

SaltyCucumber:
If you open dapp for the first time in that browser and you have more than one extension installed - multiple popups will appear for each extension asking user to authorize.
What I want is to have status for each extension, let user decide which one to enable
In case when user already previously enabled an extension, I as dapp dev can immediately call .enable method for that extension and get user's accounts and other data if needed.

packages/extension-compat-metamask/src/bundle.ts
```diff
@@ -84,6 +84,9 @@ function injectMetaMaskWeb3 (win: Web3Window): void {
         }
       };
     },
+    isEnabled: async (): Promise<boolean> => {
```
hamidra, line 87: Why this always return true?! Also seems this Metamask interface is depracated. https://docs.metamask.io/guide/ethereum-provider.html#ethereum-enable-deprecated
Do we still need to keep this up-to-date. I believe some migration is necessary if this is still intended to be used.
https://docs.metamask.io/guide/ethereum-provider.html#ethereum-enable-deprecated

SaltyCucumber: It's just a stud, I've put it to return true just in case so that it would not stop anything.

hamidra: I believe it should return false then, if that's the intention?! basically, the dapps that use this isEnable stub should always call enable(), since that has been the behavior before these changes.

SaltyCucumber: Let's say I use true/false, to determine whether I should connect immediately without user's action.
true would call .enable automatically, this should trigger a popup if done for the first time or just connect as usual if done a second or more times. If it's always false, then even if I as a user connect manually it would still show false all the time, which is not correct.

So putting true as default here makes more sense, at least I think so.

hamidra: Makes sense if that's the case.

JurbinM (approved):

Tbaut:
Interesting issue I too had to deal with eventually. iiuc you would go through the list in `window.injectedKeys` and call the `enable` for the first extension that has `isEnabled === true`, otherwise you'd ask which one the UI should connect to?

Now how would you let a user actually disconnect from an extension if say I have Pjs and Talisman? I connected Pjs first, unless I uninstall it, there's no way to set the `isEnabled` to false. We'd need to expose a `disable` here too right?

SaltyCucumber:
> Interesting issue I too had to deal with eventually. iiuc you would go through the list in `window.injectedKeys` and call the `enable` for the first extension that has `isEnabled === true`, otherwise you'd ask which one the UI should connect to?
> 
> Now how would you let a user actually disconnect from an extension if say I have Pjs and Talisman? This is something that the UI used to handle with localstorage, but would not work any more. Say I connected Pjs first, unless I uninstall it, there's no way to set the `isEnabled` to false and prevent my UI to call enable right away. We'd need to expose a `disable` here too right?

I get the full list from window.injectedWeb3, and later in UI show user which extension is connected/disconnected + automatically load accounts/data on those who are authorized, and I can do that because I know there won't be any unwanted popups.
When user opens page for the first time, the extensions are all disconnected and wait for user to choose to connect, click and only then an auth popup appears, once approved next time user won't need to do anything else with that extension and all the data will be shown (what data will be shown and how fully depends on dapp developer, currently my goal is to create a polkadot wallet selector core functionality, that any dapp dev can take and quickly onboard to the system).

Currently you can only connect yes, disconnect can be done through extension and refreshing the page, which is not ideal I agree. Let's start with connecting first.

packages/extension-compat-metamask/src/bundle.ts
```diff
@@ -84,6 +84,9 @@ function injectMetaMaskWeb3 (win: Web3Window): void {
...
+      return new Promise((resolve) => resolve(true));
```
jsidorenko, line 88:
```suggestion
      return Promise.resolve(true);
```

SaltyCucumber:
@jacogr any thoughts/comments? Can we proceed in some way, cause this functionality is important and a semi-blocker right now.

jacogr:
So what is exactly is wrong with multiple extensions asking the user to action? The user has them installed, so wish to use them. (Obviously if messages are shared, i.e. mis-configured extensions not specifying the correct environment and then getting multiple popups when signing - which seems to be the case out in the wild, that is different)

packages/extension-compat-metamask/src/bundle.ts
jacogr: This is the wrong way around - we are assuming that all extensions would prove this, which is not the case -0 it would depend on when they have been updated.

SaltyCucumber: What would be the correct way?
Currently there is no native way of knowing if extension is already connected or not. Of course we can close eyes on the multi-popup issue and just assume that most users will have only one primary extension.
There is also no way of disconnecting either, user needs to open the extension and disconnect manually, that is assuming the user knows how to do it.

jacogr: I'm saying, if we have to add `isEnabled`, it won't be available on all extensions for literally months. So, if we do make any backwards incompatible changes that may or may not be used, the absence of that function or getter should be the default.

So instead of -

```js
if (!ext.isEnabled || ext.isEnabled()) {
  // something is enabled, we had to know that default is true
}
```

we do the reverse -

```js
if (!ext.isDisabled) {
  // something is enabled
}
```

In this case it is even better if we use a getter, so `false == undefined` in JS fashion.

SaltyCucumber: Screenshot 2022-08-31 at 17 52 32
in my proof of concept code, I just used this
a simple check if method exists, if it does execute it and return result, otherwise return null
which means if I get null, I as a dapp dev can just run the .enable method by default, since .isEnabled isn't implemented in this case.

jacogr: A getter is still preferred. (Assuming this functionality is actually required)

hamidra:
@jacogr the main idea behind having this flag is to be able to enable granular wallet selection at the UX level. e.g. comparing the two behaviors(polkadot.js vs staking dashboard).
https://www.youtube.com/watch?v=WcrnU646DXc

The changes definitely should be backward compatible, so anyone who wants to pull all accounts from all extensions can use the web3Enable ,.... , but if they want to provide a granular wallet selection currently they would need to track which wallets are connected and store it in storage, but having this zero cost flag enables the dapps to query the Authorization state from the source instead of guessing it based on what is stored in session/app storage (which will be reset often). 

IMO, pulling all accounts makes sense for polkadot.js which mostly is used by savvy users, but for common users, it is better to guide them through connection so they know what are the popups and need to take action. since just popping multiple authz windows might confuse them and they might not take proper action to authorize the extensions and might get stuck in a pending state.

packages/extension-compat-metamask/src/bundle.ts
hamidra: Getter makes sense.
no strong opinion, but isDisable might create those double negative conditions e.g. if someone would need to check "not isDisabled" which might be less idiomatic and harder to follow.

fend25 (approved):

saltict:
Hi @jacogr and @SaltyCucumber ! I'm from SubWallet team.

I think this PR is really helpful to improve the UX of DApps. Wallet connection status checking and wallet interaction should be separated. Adding `isEnabled()` or `isDisabled()` is a way to do it, however, adding more detailed statuses is the best way. There are these cases to look into:
- DApp hasn't connected to the wallet
- Dapp has been accepted to connect to the wallet
- DApp is blocked by the wallet

In these cases, returning a simple boolean showing connection status is not enough. These cases have more complicated logic

Let's consider about something like that:
```typescript
{
   permission: 'not-authorized' | 'authorized' | 'rejected'
}
```

It is also easy to check if the wallet has supported this feature or not, which means better versioning 
```typescript
wallet.permission === undefined
```

Hope my suggestion can help you.

PR #1133, opened 2022-08-29
Remember last account selection for authorization

### LINKED ISSUE #1132: Save prev selection for account approvals
Currently it is a bit of a "slog" to scroll down and select accounts of each authorization. It would be better to save the previous selection and supply that as the default.

(I think this is the only reason I have not released it to stores as of yet - it just feels a tiny bit cumbersome)

Tbaut:
closes #1132

I went for an additional key in the `AccountJson` type to avoid any additional message passing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,3 +26,4 @@ package-lock.json
 _book
 docs/html
 .idea
+.vscode
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -34,14 +34,6 @@ function getSuri (seed: string, type?: KeypairType): string {
     : seed;
 }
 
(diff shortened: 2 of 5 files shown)


jacogr (commented):
Ok, need to chew through this one - I would have gone for just storing it in UI localStorage since it is purely a UI thing. (Doesn't make saving it on the back-end "wrong" just "different". And the backend storage could well be more optimal since we may be writing the same storage anyway)

Tbaut:
The main reason I went with this, is that it felt cleaner because it centralizes anything related to this auth-account lists. Having a list in the Ui for default and a list of accounts in the bg made me think that this part could be forgotten in the future. Like when an account gets deleted, the auth-account list and the default are taken car of in the same function.

jacogr (approved):
Thank you. Seems sane.

Issue #1037, opened 2022-03-29
Enhance account visibility management - authorization per account and per website


Tbaut:
It has been proposed in different ways in the past, and there has been a [grant](https://github.com/w3f/Grants-Program/blob/master/rfps/privacy-enhancement-polkadot-extension.md) for it. I'd like to tackle this in a PR, but first, I'd like to get your soft approval of the general idea and flow I'm thinking of.

#### Problem:
The eye open/close button per account works for all websites. If users want to show only account A on website A, and account B on website B, they'd have to go to the extension, make sure only account A is visible when they browse website A, or show only account B when they browse on website B. This is impracticable, and leads to the fact that users have most of their accounts visible for all website, which is a privacy threat.

#### Proposed solution:
Instead of having a mapping `website - authorization` and then `account - visibility`. I propose to have `website - authorization - account`. In practice the flow would be as follow:
- When connecting to a new website, users are prompted not only to give the authorization for this website to see their accounts, but also let them select which accounts are visible for this website
- Enhance the website management page to show how many accounts are connected, and on click which ones.
- Changing the current behavior of the eye open/close on the account overview to reflect the state for the currently visited website, and allow to change it with a single click. So as not to confuse users, and clearly show the new behavior, I would go for a different icon than the eye. Probably something like a link and broken link.
- As for the migration, I'd have to research a bit more the practicality, but if possible, I would start from scratch, and request new access for any website.

What do you think?

Tbaut:
Thanks for your input. I'm definitely for something very much different than what the grant suggests :)

I think we should have a quicker way to make an account visible to a website than having to go to website management, but this is something we can see by actually using it, or we can create shortcuts by showing the current website at the top of the list for instance.

Sounds good for your suggestions having a quick button to select all is certainly handy for some use cases.

jacogr:
We can certainly enhance from v1 - so would suggest getting the basics in-place and then make it "quick".

jonasW3F:
Thanks for the discussion and the motivation to make this happen. The RFP was stating the problem and offered some initial ideas as solution, but I am certainly open to other and more lean approaches, especially from people that know the code. As long as this offers better account handling and improves the privacy of users in the end.

With regard to this:

> Changing the current behavior of the eye open/close on the account overview to reflect the state of the currently visited website, and allow to change it with a single click. So as not to confuse users, and clearly show the new behavior, I would go for a different icon than the eye. Probably something like a link and broken link.

[...]

PR #1138, opened 2022-09-04
Add link to accounts connected to the current tab in the header

Tbaut:
- Issue: when a set of accounts are shared with a dapp, it is not easy for users to change the authorized accounts, it takes several clicks and it is error prone to search for the website manually. They'd have to go to setting > website management > find the current url > click on the number of accounts.
- Solution: when the current tab is a website that has been connected to in the past, show a "Connected" button in the header that will lead the the website management of the associated Dapp right away.

This issue has been reported to me by a user testing master, and I saw this issue coming back then here https://github.com/polkadot-js/extension/issues/1037#issuecomment-1081834987
I have tested this heavily with Brave and Firefox in light and dark mode.
The header is hidden when search is active to leave space.

Example with multi-windows browser

https://user-images.githubusercontent.com/33178835/188335115-761549a0-9fc3-480b-8b9f-eb7786d5a77e.mp4


jacogr (approved):
Cool.

I'm approving it after reading though everything multiple times, without actually running it (unlike what I normally do, aka merge is based on running). 

If I do spot issues, will log issues and submit PRs. This is one of those that seems fine, but probably has a hook-or-two that _might_ yield some small issues.

PR #1139, opened 2022-09-07
Use go-back instead of going to overview from the auth management

Tbaut:
When coming from the home (by clicking on the new "connected" button), changing the set of connected accounts would then redirect to the "Website Management" page. This is unexpected since the users came from the home directly.

Using `goBack` here fixes this unexpected behavior.

### DIFF
--- a/packages/extension-ui/src/Popup/AuthManagement/AccountManagement.tsx
+++ b/packages/extension-ui/src/Popup/AuthManagement/AccountManagement.tsx
@@ -5,9 +5,10 @@ import type { ThemeProps } from '../../types';
 
 import React, { useCallback, useContext, useEffect } from 'react';
 import { useParams } from 'react-router';
+import { useHistory } from 'react-router-dom';
 import styled from 'styled-components';
 
-import { AccountContext, ActionContext, Button } from '../../components';
+import { AccountContext, Button } from '../../components';
 import useTranslation from '../../hooks/useTranslation';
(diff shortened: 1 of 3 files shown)


jacogr:
Just for the sake of consistency, wouldn't it make more sense to extend `onAction` a bit - https://github.com/polkadot-js/extension/blob/f0a7bee95369c7a72087e6f8726bdb3ea036d200/packages/extension-ui/src/Popup/index.tsx#L81-L90

Basically allowing `..` that when found, would use `goBack`. (That way we have a built-in go back functionality which is no doubt useful elsewhere as well)

Tbaut:
Sure, I'll add it

Tbaut:
I wanted to use the `onAction('..')` in the backArrow click in the `Header` component, effectively replacing a `onAction('/')`, but because it was breaking the flow for "import account from JSON" or "connect Ledger device" that open the extension in a new tab and clears the navigation, I'll need to take care of this. I'm worried I'll break other unexpected things though.

I'll push it so you can take a look, and we can revert if deemed too risky

jacogr (approved):
Agreed. Keep this one just to this specific issue (as here), then can test other places in a different PR where it _may_ make sense.

Tbaut:
I ended up finding a middle ground, by checking the `history`, in case we can't go back, then go home. lmk what you think.

jacogr (approved):

jacogr:
Ticks all the boxes :)

PR #1147, opened 2022-09-26
change textarea for div to show the mnemonic seed

Gioyik:
_context:_
the browser stores input/textarea values in tabs into plain text files in disk to restore content in case of a browser crash, unexpected closing or session restore. Multiple wallets and signers used textarea to show the generated Mnemonic Phrase seeds, obtaining the risk of getting into a situation like [this](https://halborn.com/halborn-discovers-critical-vulnerability-affecting-crypto-wallet-browser-extensions/).

_proposal:_
show the Mnemonic Phrase seed in a non input type HTML Element to avoid leaking of phrase in the cases mentioned before. With this change, it looks the same and for the user should not represent any change, there is a TODO related to the approach to the clipboard, as I still haven't check if using the Clipboard API requires asking for the additional permission to the browser.

### DIFF
--- a/packages/extension-ui/src/Popup/CreateAccount/Mnemonic.tsx
+++ b/packages/extension-ui/src/Popup/CreateAccount/Mnemonic.tsx
@@ -12,26 +12,12 @@ interface Props {
   seed: string;
 }
 
-const onCopy = (): void => {
-  const mnemonicSeedTextElement = document.querySelector('textarea');
-
-  if (!mnemonicSeedTextElement) {
-    return;
-  }
-
-  mnemonicSeedTextElement.select();
-
-  // eslint-disable-next-line deprecation/deprecation
-  document.execCommand('copy');
-};
-
 function Mnemonic ({ onNextStep, seed }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const [isMnemonicSaved, setIsMnemonicSaved] = useState(false);
   const { show } = useToast();
 
   const _onCopy = useCallback((): void => {
-    onCopy();
     show(t<string>('Copied'));
   }, [show, t]);
 
--- a/packages/extension-ui/src/Popup/CreateAccount/CreateAccount.spec.tsx
+++ b/packages/extension-ui/src/Popup/CreateAccount/CreateAccount.spec.tsx
@@ -72,8 +72,8 @@ describe('Create Account', () => {
   });
 
   describe('Phase 1', () => {
-    it('shows seed phrase in textarea', () => {
-      expect(wrapper.find('textarea').text()).toBe(exampleAccount.seed);
+    it('shows seed phrase in a span inside a div', () => {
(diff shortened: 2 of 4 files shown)


jacogr:
The approach sounds sane - will take a look.

jacogr:
Failing both linting & tests. (Running `yarn lint` locally should address)

Tbaut:
@Gioyik Any chance you can fix the nit please so that we can merge?

Tbaut:
I don't have a strong opinion on it, so I went ahead and checked what has been used in other implementations such as the `@polkadot/react-identicon` package. We use an external package there called `react-copy-to-clipboard`, [here](https://github.com/polkadot-js/ui/blob/master/packages/react-identicon/src/Identicon.tsx#L8) which is itself using [`copy-to-clipboard`](https://www.npmjs.com/package/copy-to-clipboard). So I'd go for this :shrug:

packages/extension-ui/src/Popup/CreateAccount/Mnemonic.tsx
```diff
@@ -12,26 +12,13 @@ interface Props {
...
-};
-
 function Mnemonic ({ onNextStep, seed }: Props): React.ReactElement<Props> {
   const { t } = useTranslation();
   const [isMnemonicSaved, setIsMnemonicSaved] = useState(false);
   const { show } = useToast();
 
   const _onCopy = useCallback((): void => {
-    onCopy();
+    setIsMnemonicSaved(true);
```
Tbaut, line 21: copying doesn't mean that the users have saved the mnemonic safely. I'd suggest to keep the manual checkbox as it was.

```suggestion
```

Tbaut (changes requested):
Thank you. It looks good beside one thing that was added and that I disagree with. I think users should read and manually check the box, since copying doesn't mean that it is saved safely.

Tbaut (approved):
Thanks, looks good. I tested on Brave

jacogr (changes requested):
All good. Happy to merge as-is.

(Could you just merge in master one last time to resolve conflicts, once done will approve)

Tbaut:
Could you check again @Gioyik, the lint and build failed.

Gioyik:
@Tbaut @jacogr fixed. Sorry, I forgot to check it builds before pushing. Should be good now.

packages/extension-ui/src/Popup/CreateAccount/Mnemonic.tsx
Tbaut: This is still here BTW. Please remove it.

Tbaut:
The `setIsMnemonicSaved(true);` slipped in again unfortunately. Please remove it.

Gioyik:
@Tbaut done

jacogr (approved):
Thank you.

PR #1155, opened 2022-10-13
Allow QR signing and add error when the account has no genesis

Tbaut:
This PR reverts https://github.com/polkadot-js/extension/pull/1053 now that Signer supports raw bytes signing again thanks to  https://github.com/paritytech/parity-signer/pull/1332 

Small bonus, I added an error message in case the account has no `genesisHash` associated. This is mandatory for QR generation and is a common source of confusion. Before having this message, the screen would only show the bytes, with no signing button or QR code, leaving users with no clue about what's going on. 

I tested it with Signer `master` branch, Polkassembly and the Signing tool of apps.

image
image
image

### DIFF
--- a/packages/extension-ui/src/Popup/Signing/Request/index.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Request/index.tsx
@@ -6,7 +6,7 @@ import type { ExtrinsicPayload } from '@polkadot/types/interfaces';
 import type { SignerPayloadJSON, SignerPayloadRaw } from '@polkadot/types/types';
 import type { HexString } from '@polkadot/util/types';
 
-import React, { useCallback, useContext, useEffect, useState } from 'react';
+import React, { useCallback, useContext, useEffect, useMemo, useState } from 'react';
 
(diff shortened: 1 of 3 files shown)


packages/extension-ui/src/Popup/Signing/Request/index.tsx
```diff
@@ -126,14 +128,14 @@ export default function Request ({ account: { accountIndex, addressOffset, isExt
...
           isFirst={isFirst}
-          isSignable
           setError={setError}
           signId={signId}
         />
       </>
     );
   } else if (hexBytes !== null) {
     const { address, data } = request.payload as SignerPayloadRaw;
+    const account = accounts.find((account) => decodeAddress(account.address).toString() === decodeAddress(address).toString());
```
jacogr, line 138: This is quite a heady operation, since it uses decodeAddress (twice). This should be memo-ized.

jacogr (approved):
Thank you!

mustermeiszer:
Thanks for being so fast @Tbaut! 

@jacogr any estimates when this will hit the public chrome extension?

PR #1156, opened 2022-10-16
Fixed loading error of dev version on Firefox

JasperAlexander:
Fixed [this](https://substrate.stackexchange.com/questions/5371/index-html-file-not-found-on-firefox-after-building-dev-version-of-polkadot-js-e) this problem. You should load the index.html file on Firefox instead of the manifest.json file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Steps to build the extension and view your changes in a browser:
   - Firefox:
     - go to `about:debugging#addons`
     - check "Enable add-on debugging"
-    - click on "Load Temporary Add-on" and point to `packages/extension/build/manifest.json`
+    - click on "Load Temporary Add-on" and point to `packages/extension/build/index.html`
     - if developing, after making changes - reload the extension
 3. When visiting `https://polkadot.js.org/apps/` it will inject the extension
 


jacogr:
Which version of FF - weirdly enough, the `manifest.json` works as expected for me.

@Tbaut Which approach do you normally take here?

Tbaut:
I have definitely always pointed at the manifest. I think the best approach is to see what the Firefox documentation says that any file should work cf https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Your_first_WebExtension#installing

JasperAlexander:
Very strange! I've cloned the repo again and executed the same steps (yarn install, yarn build). Now when I load the manifest file of the newly cloned repo I'm having the same problem (file not found), but when I load the manifest file from yesterday it does work. Maybe a Firefox bug? Can someone try the same steps as I did? I've Firefox version 105.0.3.

jacogr:
Which OS? (I am on Mac, Tibaut on Linux … unless he swapped recently)

My FF is also latest, updated this morning.

JasperAlexander:
> Which OS? (I am on Mac, Tibaut on Linux … unless he swapped recently)
> 
> 
> 
> My FF is also latest, updated this morning. 

Latest MacOS. Have you cloned the repository?

jacogr:
Fresh clone, all good, loads the `packages/extension/build/mainfest.json` as expected.

So it seems _very_ suspect that somewhere this doesn't operate as expected. Unsure as to why.

PR #1242, opened 2023-03-29
Resize for popup for consistent size on Windows and Mac OS

F-OBrien:
Due to the differences in the way that Windows and Mac OS renders the window frames and borders it results in inconsistent popup sizes which obscures some information. e.g.  the "Ask again later" option as shown below for Windows. This only happens when a popup is created and not when clicking on the extension icon. `chrome.windows.create` doesn't have a clean way to handle both OS's. This PR is a hacky way to try to handle both OS's.

Original `chrome.windows.create` outer dimensions = 560x621 (from `POPUP_WINDOW_OPTS`)
Resulting Windows popup inner dimension = 544x582
image

After resize to 560x600 inner dimension
image

### DIFF
--- a/packages/extension-ui/src/createView.tsx
+++ b/packages/extension-ui/src/createView.tsx
@@ -10,6 +10,24 @@ import { HashRouter } from 'react-router-dom';
 import { View } from './components/index.js';
 
 export default function createView (Entry: React.ComponentType, rootId = 'root'): void {
+  const targetInnerWidth = 560;
+  const targetInnerHeight = 600;
+
+  // Popup window size is initially set via chrome.windows.create, however
+  // it will result in different inner dimensions on Windows and Mac OS.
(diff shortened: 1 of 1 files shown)


Tbaut:
I tested it on Linux and the hack doesn't seem to work. Here's a QR code request: 
image

jacogr:
... These window sizes between browsers and platforms has been a huge PITA since the start...

F-OBrien:
That's disappointing. Well it seems to work with Chrome at least if you wanted to consider it as a slight improvement over the current implementation.

jacogr:
I'll defer to @Tbaut on this one since (like he mentioned above), he spent a lot of time on this in the past.

My general feeling is that - if it is better in some cases and no worse in others (e.g. FF), it is certainly worth a merge. 

TL;DR As long as it doesn't make anything worse, I'm ok with it.

PR #1308, opened 2024-01-19
Fix: Prevent authorization request from incorrect origin due to chrome pre-rendering fixes #1293 

### LINKED ISSUE #1293: Incorrect origin URL shown in the pannel
- [x] Bug report
- [x] Browser (Google Chome  119.0.6045.123)

While I was browsing a GitHub page, a pop-up window unexpectedly appeared. This occurred possibly after switching from a tab displaying the `Phat Contract UI`, although I'm not certain of the exact sequence of my actions. The pop-up is captured in the following image: image.

I suspect this could potentially indicate a security vulnerability.

F-OBrien:
We cannot rely on `sender.tab.url` when chrome preloads and prerenders pages as it will not match the url of the site being loaded. This PR changes the primary from to `sender.url` and only uses `sender.tab.url` if `sender.url` is not available.

See discussion in #1293 

Edit:
Also PR #1278 introduced a bug by adding
```
  if (!sender) {
    throw new Error('Unable to extract message sender');
  }
```
in the case of [getActiveTabs](https://github.com/polkadot-js/extension/blob/128738181afb5796e326632c60f861e278e6481b/packages/extension/src/background.ts#L32) no port is passed causing this to throw an error preventing the extension from initializing.

This PR fixes that by only throwing `if (!isExtension && !sender)`

### DIFF
--- a/packages/extension-base/src/background/handlers/index.ts
+++ b/packages/extension-base/src/background/handlers/index.ts
@@ -22,13 +22,13 @@ export default function handler<TMessageType extends MessageTypes> ({ id, messag
   const isExtension = !port || port?.name === extensionPortName;
   const sender = port?.sender;
 
-  if (!sender) {
-    throw new Error('Unable to extract message sender');
+  if (!isExtension && !sender) {
+      throw new Error('Unable to extract message sender');
   }
 
   const from = isExtension
     ? 'extension'
-    : (sender.tab?.url) || sender.url || '<unknown>';
+    : sender?.url || sender?.tab?.url ||  '<unknown>';
   const source = `${from}: ${id}: ${message}`;
 
   console.log(` [in] ${source}`); // :: ${JSON.stringify(request)}`);
--- a/packages/extension-base/src/background/handlers/index.ts
+++ b/packages/extension-base/src/background/handlers/index.ts
@@ -23,12 +23,12 @@ export default function handler<TMessageType extends MessageTypes> ({ id, messag
   const sender = port?.sender;
 
   if (!isExtension && !sender) {
-      throw new Error('Unable to extract message sender');
+    throw new Error('Unable to extract message sender');
   }
 
   const from = isExtension
     ? 'extension'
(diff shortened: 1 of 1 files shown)


Jess25051980 (approved):

TarikGul:
If you can pull in master that would be amazing so the CI can run again, and I will do a proper review of this PR.

TarikGul:
@Tbaut Could you also check this out when you have a chance?

F-OBrien:
@TarikGul I've rebased rebased on master as requested. 
FYI Talisman also applied a similar fix today https://github.com/TalismanSociety/talisman/pull/1349

Tbaut:
Sure, I'm not familiar with the issue (despite having seen and followed the linked one). I'll take a look.

F-OBrien:
> @F-OBrien you mention in the last comment: [#1293 (comment)](https://github.com/polkadot-js/extension/issues/1293#issuecomment-1900267262) that it's better to not support iframes (I agree) and then say
> 
> > sender.documentLifecycle === 'prerender' to use the sender.url as the from
> 
> But this is not what you have done here. Is there a reason for it?

@Tbaut It's been a while since I looked at this. From what I recall I believe I concluded that in all cases using `sender.url` should be the preferred solution and not only if `sender.documentLifecycle === 'prerender'` so didn't see a point in an explicit check for it. `sender.tab.url` is the cause of the incorrect origin issue and also gives rise to my concern related to iframes.

I did consider fully removing `sender.tab.url` like talisman has done but I was not familiar enough with chrome extensions or the reason why the tab url was initially chosen when developing the extension to know if these is a specific edge case where you would have a `sender.tab.url` and not a `sender.url`. hence why I just swapped them rather than removing the tab url.

Tbaut (approved):

Tbaut:
Thank you so much for the research and the fix :pray:

TarikGul (approved):
Ahh thanks for the explanation 🙏 , and PR. LGTM

Issue #1314, opened 2024-02-24
 Unable to use the `ChargeAssetTxPayment` signed-extension on Westend Asset Hub


josepot:
* **I'm submitting a Bug report**

* **What is the current behavior and expected behavior?**

As I'm sure that you already know: It is possible on Polkadot-SDK runtimes that support the `ChargeAssetTxPayment` signed-extension to pay the transaction fees using an asset that can be swapped for the native currency of the chain.

In some runtimes the `ChargeAssetTxPayment` signed-extension is configured to use a `u32` for indicating a "local" `asset_id` present on that particular chain, that is currently the case on "Polkadot Asset Hub".

However, the `ChargeAssetTxPayment ` signed-extension could also be configured to use (for instance) an `XcmV3MultiLocation` path indicating the asset. This is currently the case of "Westend Asset Hub", and -unless I'm mistaken- that will also be the case for "Polkadot Asset Hub" at some point in the not so distant future.

The issue is that the PJS extension crashes on dApps that try to leverage this signed-extension when the runtime is configured to use an `asset_id` different than `Option(u32)`:

![image](https://github.com/polkadot-js/extension/assets/8620144/656b5b9c-231d-4b3e-a62f-6cae252a29c1)

This happens even when the extension has been updated with latest metadata. Also, the "Back to Home" button does nothing, and the only way to get the extension functional again is to disable it and to re-enable it.

Therefore, it seems that the PJS extension assumes that the `asset_id` field of the `ChargeAssetTxPayment` signed-extension will always be an `Option(u32)`, which is not a safe assumption. The extension should, instead, check on the metadata what's the actual shape of this field and present and sign the data that's receiving from the dApp according to the type that's defined on the metadata.

FWIW that's what Polkadot-API does. Meaning that when a developer is trying to leverage this signed-extension on "Westend Asset Hub" [the types look like this](https://github.com/josepot/westend-asset-hub-test-papi/blob/47c28355e66f3a67e9449d18b638135e44564123/src/App.tsx#L111-L117), and the types of that same field on "Polkadot Asset Hub" are a plain `number`, while on chains that don't support the `ChargeAssetTxPayment` the types won't even allow the developer to pass the `asset` field.

I came across this issue thanks to @jameshih who asked me to "translate" the [following POC dApp](https://github.com/jameshih/westmint-test-polkadot-js) that he had built using PJS to Polkadot-API. What this POC dApp does is: 

[...]

PR #1331, opened 2024-03-21
feat(extension-ui): extrinsic asset id

ryanleecode:
Fixes: https://github.com/polkadot-js/extension/issues/1314

Adds `assetId` in the extrinsic UI. Assumes the assetId is an XCMV2 junction object with the junctions `PalletInstance` and `GeneralIndex`. Otherwise it won't render and neither will it crash because any errors are swallowed to `null`.

image


marshacb (approved):
lgtm!

Some things to noted from when we spoke:

* supporting the`X2` junction may be limiting and could be expanded to support other location junctions but if this is tout of scope for this fix another issue can be created
* PJS apps doesn't have the `assetId` appear in the extension UI due to the assetId not being a part of the payload and would require additional changes

ryanleecode:
> lgtm!
> 
> Some things to note from when we spoke:
> 
> * works when sending an X2 pool asset location as the optional `assetId` (e.g. `{parents: 0, interior:{X2:[{PalletInstance:50},{GeneralIndex:41}]}}`
> * supporting only the`X2` junction may be limiting and could be expanded to support other location junctions (if it turns out to be necessary) but if this is out of scope for this fix another issue can be created
> * PJS apps doesn't have the `assetId` appear in the extension UI due to the assetId value not being a part of the `payload` and would require additional changes

We should open an issue in polkadot-sdk to get that error changed as well. "Invalid Transaction: Payment" is far too vague, and doesn't tell you that there is no liquidity pool for this asset.

packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
```diff
@@ -26,6 +26,32 @@ interface Props {
...
+    'PalletInstance' in assetId.interior.X2[0] &&
+    'GeneralIndex' in assetId.interior.X2[1];
+}
+
+function decodeAssetId (assetId: unknown) {
+  if (!isAssetId(assetId)) {
+    return null;
+  }
+
+  try {
```
TarikGul, line 48: For the return value of the AssetId I think we can just return the whole `MultiLocation`, and shouldn't extrapolate any specific, this will keep it true to the original input. 

cc: @IkerAlus Thoughts?

TarikGul (changes requested):
Couple discussion points on the following. Thank so much for the PR.

packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
IkerAlus: Why are we limiting this to only asset IDs as XCMV2 junction objects with the junctions `PalletInstance` and `GeneralIndex`? Is there a limitation I am not aware of?

From UX perspective, indeed, most of the cases of users paying fees in other tokens beyond will have an asset ID in the mentioned form, but actually any asset with an existing pool against DOT (asset conversion pallet) will be able to do this, for example any foreign asset sent to AH. 

The general solution will be to allow any `Multilocation` as asset ID.

ryanleecode: We thought of this but I think this XCMV2 object covers 99% of usecases. This change doesn't affect the actual functionality, so if the user provides say and interior with length 3/4/5/6/7/8, it will still go through. This function is just to display it in the UI. 

For the other types, its ambigious how it would be displayed in the UI. should we just assume the last element in the array is always `GeneralIndex` and it is the asset id?

packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
```diff
@@ -26,6 +26,32 @@ interface Props {
...
 }
 
+interface AssetId { parents: number, interior: { 'X2': [{ PalletInstance: string }, { GeneralIndex: string }] } }
+
+function isAssetId (assetId: unknown): assetId is AssetId {
+  if (!assetId || typeof assetId !== 'object') {
+    return false;
+  }
+
+  return 'parents' in assetId && 'interior' in assetId &&
```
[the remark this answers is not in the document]
ryanleecode: Yeah this is an option, we can just use nullish optional chaining

packages/extension-ui/src/Popup/Signing/Extrinsic.tsx
IkerAlus: I see, thanks! Indeed, it makes sense to only show these cases in the UI, particularly considering it won't crash and simply not show the general case.

TarikGul: Isn't the whole multilocation considered the Asset though? So should we show the whole multilocation?

ryanleecode:
@TarikGul @IkerAlus How does this look

image

TarikGul:
@ryanleecode Looks really solid, nice job!

For the Dropdown could we do something like:

```
<tr>
    <td className='label'>{t('method')}</td>
    <td className='data'>
        <details>
            <summary>{'{...}'}</summary>
            <pre>{someValueForJSOn}</pre>
        </details>
    </td>
</tr>
```

ryanleecode:
> @ryanleecode Looks really solid, nice job!
> 
> For the expendable could we do something like:
> 
> ```
> <tr>
>     <td className='label'>{t('asset')}</td>
>     <td className='data'>
>         <details>
>             <summary>{'{...}'}</summary>
>             <pre>{someValueForJSOn}</pre>
>         </details>
>     </td>
> </tr>
> ```

the thing i dont like about this is you need that extra click to see the actual asset id where as before it was `assetId: 41` so you could easily tell what the asset was and sign it.

image

even worse is if the xcm is really big u might even need to scroll down

image

TarikGul:
> the thing i don't like about this is you need that extra click to see the actual asset id where as before it was assetId: 41 so you could easily tell what the asset was and sign it.

I think there is two key points here that i see:

- The user should have visibility into the exact information they inputted to build the payload. A displayed `u32` is much different than a `MultiLocation` in terms of what the user inputs.
    - The `GeneralIndex` is only part of the information that characterizes the asset. For example: The value passed into the `PalletInstance` is just as important as the `GeneralIndex`, as well as the `parents`. So being able to verify each field of the Multilocation would be super useful for the user.

ryanleecode:
> > the thing i don't like about this is you need that extra click to see the actual asset id where as before it was assetId: 41 so you could easily tell what the asset was and sign it.
> 
> I think there is two key points here that i see:
> 
> * The user should have visibility into the exact information they inputted to build the payload. A displayed `u32` is much different than a `MultiLocation` in terms of what the user inputs.
>   
>   * The `GeneralIndex` is only part of the information that characterizes the asset. For example: The value passed into the `PalletInstance` is just as important as the `GeneralIndex`, as well as the `parents`. So being able to verify each field of the Multilocation would be super useful for the user.

Ideally there should be a tool out there to parse the MultiLocation into its [written format](https://github.com/paritytech/xcm-format?tab=readme-ov-file#written-format). I haven't found one such tool so I've opened an issue in xcm-tools to request this: https://github.com/paraspell/xcm-tools/issues/188

IkerAlus:
> Ideally there should be a tool out there to parse the MultiLocation into its [written format](https://github.com/paritytech/xcm-format?tab=readme-ov-file#written-format). I haven't found one such tool so I've opened an issue in xcm-tools to request this: [paraspell/xcm-tools#188](https://github.com/paraspell/xcm-tools/issues/188)

I agree it will be nicer to show the location in the _written format_. Assuming we go this way, will the interface show the whole location at first sight? or will the user still need to do the extra click?

Dioliode (approved):

TarikGul:
So I looked into the `xcm-tools` and I am weary of adding another dependency, example comment https://github.com/polkadot-js/apps/pull/10078#issuecomment-1851375493.

One main thing is xcm-tools uses polkadot-js as direct dependency meaning the package will always need to be in sync with our updates which brings further maintenance, and potential complications. @Tbaut Any thoughts about the design.

My thought is we just keep the raw JSON as the displayed output with the dropdown, and avoid any additional maintenance of deps in the future.

ryanleecode:
> So I looked into the `xcm-tools` and I am weary of adding another dependency, example comment https://github.com/polkadot-js/apps/pull/10078#issuecomment-1851375493.
> 
> One main thing is xcm-tools uses polkadot-js as direct dependency meaning the package will always need to be in sync with our updates which brings further maintenance, and potential complications. @Tbaut Any thoughts about the design.
> 
> My thought is we just keep the raw JSON as the displayed output with the dropdown, and avoid any additional maintenance of deps in the future. 

Its possible they make it a standalone package (they already have a monorepo structure in place). in theory the package should require no dependencies on its own, since the XCM format is known.

ryanleecode:
@TarikGul Its in their pipeline for early April. 

https://github.com/paraspell/xcm-tools/issues/188#issuecomment-2025319103

Shall we wait or get this merged with `(...)` for now and open another issue to followup

TarikGul:
> @TarikGul Its in their pipeline for early April.
> 
> [paraspell/xcm-tools#188 (comment)](https://github.com/paraspell/xcm-tools/issues/188#issuecomment-2025319103)
> 
> Shall we wait or get this merged with `(...)` for now and open another issue to followup

Happy to get this in, and we can create a tracking issue to follow this up.

TarikGul (approved):

Dioliode (approved):

PR #1336, opened 2024-03-23
docs: update readme

ltfschoen:


Dioliode (approved):

Tbaut:
These are very specific changes to your needs. There is no need for changing the name, or install nvm.

Dioliode (approved):

ltfschoen:
> These are very specific changes to your needs. There is no need for changing the name, or install nvm.

i've updated it so use of nvm is optional and removed other stuff that isn't essential

README.md
```diff
@@ -18,19 +18,23 @@ Find out more about how to use the extension as a Dapp developper, cookbook, as
...
   - Chrome:
     - go to `chrome://extensions/`
     - ensure you have the Development flag set
     - "Load unpacked" and point to `packages/extension/build`
     - if developing, after making changes - refresh the extension
   - Firefox:
-    - go to `about:debugging#addons`
-    - check "Enable add-on debugging"
-    - click on "Load Temporary Add-on" and point to `packages/extension/build/manifest.json`
+    - go to "This Firefox" at `about:debugging#addons`
```
TarikGul, line 31: What does `"This Firefox"` mean?

ltfschoen: it's a menu item in latest Firefox 124.0.1 as shown below. if i go to `about:debugging#addons` it redirects me to `about:debugging#/runtime/this-firefox`

Screenshot 2024-03-25 at 8 01 44 am

README.md
```diff
@@ -18,19 +18,23 @@ Find out more about how to use the extension as a Dapp developper, cookbook, as
...
+    - check "Enable add-on debugging" (if available)
```
Tbaut, line 32: this is not present in new versions, let's remove it then.
```suggestion
```

README.md
Tbaut: This is the default tab when going to `about:debugging#addons`, so I suggest to remove this mention as it may change.
```suggestion
    - go to `about:debugging#addons`
```

README.md
```diff
@@ -18,19 +18,23 @@ Find out more about how to use the extension as a Dapp developper, cookbook, as
...
+    - go to "Temporary Extensions" and click on "Load Temporary Add-on" and point to `packages/extension/build/manifest.json`
     - if developing, after making changes - reload the extension
-3. When visiting `https://polkadot.js.org/apps/` it will inject the extension
+    - edit the extension settings to "Run in Private Windows" if necessary
+    - click the puzzle icon, then click the gear icon next to the "polkadot{.js} extension (development only)" extension and choose "Pin to Toolbar"
```
Tbaut, line 36: This is really not needed either and can change in the future. I'd remove.

```suggestion
```

PR #1340, opened 2024-03-27
Upgrade `web3` dependency in `extension-compat-metamask`

gdethier:
We are relying on `extension-compat-metamask` in our project to interact with MetaMask. However, the old `web3` dependency triggers dependabot alerts that can't be solved without upgrading.

We built a [custom package](https://www.npmjs.com/package/@logion/extension-compat-metamask) from the code in this PR (happy to remove it as soon as a new release contains this change) for testing the upgrade and everything seems to work as expected.

### DIFF
--- a/packages/extension-compat-metamask/package.json
+++ b/packages/extension-compat-metamask/package.json
@@ -26,7 +26,7 @@
     "@polkadot/types": "^10.13.1",
     "@polkadot/util": "^12.6.2",
     "tslib": "^2.6.2",
-    "web3": "^1.10.3"
+    "web3": "^4.7.0"
   },
   "peerDependencies": {
     "@polkadot/api": "*",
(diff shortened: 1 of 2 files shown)


TarikGul (changes requested):
Thanks for the patience. Conflicts just need to be resolved.

gdethier:
@TarikGul could you approve the merge? Otherwise, there will be conflicts again...

TarikGul:
@gdethier Yes I will merge this today, and dont worry no other PR will go in before this. I just need to double check the changes between the web3 versions.

gdethier:
Great, thanks!

TarikGul (approved):
So AFAIK the web3 package is only used for the `utils.sha3` signature, which has only seen some benefits in the CHANGELOG notes from what I can tell. LGTM

PR #1345, opened 2024-04-15
feat: display asset id in xcm format

ryanleecode:
Follow up to: https://github.com/polkadot-js/extension/pull/1331

Displays the asset id in XCM format if it can be parsed. 

image

Note you can still expand the box to see the whole thing.

image


TarikGul (approved):
Looks great. Thanks so much

PR #1358, opened 2024-05-02
Enable raw payloads to be signed in extension

### LINKED ISSUE #1357: wrong signRaw signature for polkadotjs due to u8aWrapBytes that forced adding  <Bytes>...<Bytes> to "sign" function
- [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

As a developer when I want to sign a `RAW` hex payload, this payload must not be manipulated by adding any [additional data like this <Bytes> wrapper](https://github.com/polkadot-js/extension/blob/d180db583e107090bfa2ae16a102af43b9a17fcd/packages/extension-base/src/background/RequestBytesSign.ts#L23), That why it is named `signRaw`. 

Sorry, but WDF, it was a terrible idea to wrap this with <Bytes> tag in the signer, that manipulation must be done on the Dapp side, not on the signer side. the change was introduced by https://github.com/polkadot-js/extension/commit/e4ce268b1cad5e39e75a2195e3aa6d0344de7745

* **What is the motivation for changing the behavior?**

I'm crafting a raw substrate transaction, and the payload must be RAW and not manipulated given that it can not be verified for the blockchain node.

### Solution

To avoid break changes to anyone using that `signRaw` with `bytes`, I recommend including an extra `type` called 'raw' in the SignRawPayload, then if the type is `raw` no manipulation must be done.

  - Version:
  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [x] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [x] Other

S0c5:
this PR closes #1357 by adding a type 'raw' signature that enables developers to sign raw payloads using the 'signRaw' method in the extension and avoid forced data manipulation in the signing process.

### DIFF
--- a/packages/extension-base/src/background/RequestBytesSign.ts
+++ b/packages/extension-base/src/background/RequestBytesSign.ts
@@ -7,7 +7,7 @@ import type { SignerPayloadRaw } from '@polkadot/types/types';
 import type { HexString } from '@polkadot/util/types';
 import type { RequestSign } from './types.js';
 
-import { u8aToHex, u8aWrapBytes } from '@polkadot/util';
+import { u8aToHex, u8aToU8a, u8aWrapBytes } from '@polkadot/util';
 
 export default class RequestBytesSign implements RequestSign {
   public readonly payload: SignerPayloadRaw;
@@ -20,7 +20,7 @@ export default class RequestBytesSign implements RequestSign {
     return {
       signature: u8aToHex(
         pair.sign(
-          u8aWrapBytes(this.payload.data)
+          ['raw'].includes(this.payload.type) ? u8aToU8a(this.payload.data) : u8aWrapBytes(this.payload.data)
         )
       )
     };


Tbaut (changes requested):
What prevents from letting users sign a TX without showing them they are signing a TX? 

What prevents you from wrapping your payload with <bytes></bytes>?

This is a security measure that should not be removed. Also it's been there for long enough that changing this behavior would actually break Dapps :)

S0c5:
@Tbaut thanks for such a quick response!!!

## What prevents them from letting users sign a TX without showing them they are signing a TX?

I suggest a change in the UI that warns clearly and directly to users that they're about to perform a low-level operation with an unknown consequence, something:

"STOP THERE: You're about to sign an unknown TX, make sure you trust your origin, given it can be without all your funds"

`what do you think?`

## What prevents you from wrapping your payload with <bytes></bytes> which is easy enough ?

Given I'm crafting a raw substrate transaction with a custom tool called https://github.com/virto-network/virto-sdk/tree/main/sube and using a cross-platform signing library called lib wallet (https://github.com/virto-network/virto-sdk/tree/main/libwallet) which one of the backends we're developing is Polkadot js, we need to sign raw payloads in order to complete the operation and avoid the BadProof error when submiting.

Tbaut:
The security issue is about having a malicious Dapp. Having a warning on the Dapp doesn't make sense, if it's malicious there'll obviously be no warning.

To sign in to some websites, e.g Polkassembly, they let users sign a non legible random number. This is bad practice and they should let them sign some text containing a random number, but that's what they do. Say Polkassembly is hacked tomorrow and instead of a random number, it's a substrate TX that transfer all the user funds to the attacker. Well users will blind sign, because the extension won't try to decode the TX since it's a raw sign. It is absolutely essential to différenciant raw signing from transaction signing.

Regarding your use case, I don't get why you need raw signing. Why don't you let users sign the TX as a TX?

PR #1382, opened 2024-06-28
Feature/select web3 extension

AleksandrRadik:
Hi there guys! I working on one of Web3 app for staking using dApp, and we wanna add possibility for staking Polkadot on our validators for user with there web3 wallets. But I meet some problem:
Imagine that user have two extension which can injected in dApp (Polkadot.js & SubWallet for example) and when we call web3Enable func it start authorization to all of them, it's some not clear for user (why you need all my extension??) and some harder to develop for developers. So, I propose add new func which return all available extensions for developer and he can shown list for user, and user can choose one only. And of course add new args for web3Enable method for enable only one of extenstion. 
Hope you accept this, thank's for your attention, time, and work, Polkadot is amazing project and we belive in big success of it in future))

### DIFF
--- a/packages/extension-dapp/src/bundle.ts
+++ b/packages/extension-dapp/src/bundle.ts
@@ -1,7 +1,7 @@
 // Copyright 2019-2024 @polkadot/extension-dapp authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
-import type { InjectedAccount, InjectedAccountWithMeta, InjectedExtension, InjectedProviderWithMeta, InjectedWindow, ProviderList, Unsubcall, Web3AccountsOptions } from '@polkadot/extension-inject/types';
+import type { InjectedAccount, InjectedAccountWithMeta, InjectedExtension, InjectedProviderWithMeta, InjectedWindow, InjectedWindowProvider, ProviderList, Unsubcall, Web3AccountsOptions } from '@polkadot/extension-inject/types';
 
 import { isPromise, objectSpread, u8aEq } from '@polkadot/util';
 import { decodeAddress, encodeAddress } from '@polkadot/util-crypto';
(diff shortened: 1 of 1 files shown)


Tbaut:
This is already possible, since front-end devs can read the extensions available and call `enable` for the selected one. There are also tools such as https://github.com/TalismanSociety/talisman-connect/tree/6ff1e4bf1dec2c957e7ef80d4bf4381aafe3bc70/packages/connect-wallets that make this even simpler.

AleksandrRadik:
Thanks for you reply! @Tbaut 
What you mean front-dev read and enable?
Just call window.InjectedWeb3 and there in object call enable method? Oh there are exist others methods? Because in this way, webEnablePromise object will undefined and all other methods will be unavailable?

Tbaut:
The `window.InjectedWeb3` is an object with injected extensions by name. Each entry has an `enable` function.
What the package I linked proposes to do is create a modal with a list of the available extensions, allows to download them if not present, and if users have certain extensions available, then enable just the one that users select.
image

AleksandrRadik:
Okay, I got it. 
As I understand, if I wanna exclude multiple popup for auth to few extenstion, I need to write my own logic working with windows.InjectedWeb3 object?
Cause If i follow your instruction, when I enable only one wallet myself it's not setting web3EnablePromise object which needed in all other's methods like web3Accounts(), web3FromSource() etc.?
image
P.S. I check TalismanConnect package it's look good, and I think i can reuse this, cause they write there own logic for working with window.InjectedWeb3 object, they just have list of supported extnestion is good. Now I wanna understand how I can use @polkadot/extension-dapp

Tbaut:
Here are some more info regarding writing your own logic: https://github.com/polkadot-js/extension/issues/1391#issuecomment-2228211034

PR #1383, opened 2024-07-02
Adjust ui imports for deterministic bundling

TarikGul:
This PR works to adjust the imports for modules part of the `@polkadot/ui` library. The reason for this is because when firefox compares the zipped libraries we create for them there are inherent differences. These diffs come from the fact that when an import is like:

`import foo from 'bar'`

the bundled code points to absolute paths in the local file system.

Instead we want to ensure its like:

`import { foo } from 'bar'`

so that it can be deterministic and the build is the same for all machines.

Note: We were a bit lucky the ui library exported everything in multiple ways or else we would have had to make a release there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,22 @@ Find out more about how to use the extension as a Dapp developper, cookbook, as 
 2. Install dependencies via `yarn install`
 3. Build all packages via `yarn build`
   - The `/packages/extension/build` will contain the exact code used in the add-on
-4. To regenerate the dst, and src compressed files run: `yarn build:zip` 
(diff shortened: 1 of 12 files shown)


Tbaut (approved):
amazing news if this is "just" what preventing them from having a reproducible build. Thank you

TarikGul:
> amazing news if this is "just" what preventing them from having a reproducible build. Thank you

Yea, it was a little PITA to track down but this should be the last nit to take care of for the Firefox review process :)

I will merge it soon, just finishing up some last peace of mind checks.

PR #1385, opened 2024-07-05
Update XCM Analyser to v1.3.0 - fix for xcm v4

dudo50:
This PR features an update to XCM-Analyser to v1.3.0. This version contains a fix for the latest XCM version - v4 where x1 multilocations are no longer considered as objects "{}" but as arrays "[]" to remain consistent with other multilocations.

With kind regards
Team ParaSpell


TarikGul (approved):
Thank you!

Issue #1386, opened 2024-07-08
moduleResolution: Bundler


ryanleecode:
Im not able to import some types due to the `moduleResolution` being on `node` still. Is it possible to set this to `Bundler`?

PR #1387, opened 2024-07-08
chore: update module resolution to bundler

ryanleecode:
Fixes: #1386

- Error boundary had a type export error so had to refactor it to functional component.
- ~~adds jest/@types/jest to resolve compiler error after node resolution change.~~
- Replace usages of `/// <reference types="@polkadot/dev-test/globals" />` with `import '@polkadot/dev-test/globals.d.ts';` and adds `'import/extensions': 'off'` to eslint config. This is required to resolve type issues.

---

`import/extensions` is also a useless eslint rule and will be superseded by native node esm resolution. see https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/extensions.md#when-not-to-use-it

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,8 +39,10 @@
   },
   "devDependencies": {
     "@polkadot/dev": "^0.79.3",
+    "@types/jest": "^29.5.12",
     "@types/node": "^20.10.5",
     "i18next-scanner": "^4.4.0",
+    "jest": "^29.7.0",
     "sinon-chrome": "^3.0.1"
   },
   "resolutions": {
--- a/package.json
+++ b/package.json
@@ -39,10 +39,8 @@
   },
   "devDependencies": {
     "@polkadot/dev": "^0.79.3",
-    "@types/jest": "^29.5.12",
     "@types/node": "^20.10.5",
     "i18next-scanner": "^4.4.0",
-    "jest": "^29.7.0",
     "sinon-chrome": "^3.0.1"
   },
   "resolutions": {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -4,5 +4,10 @@
 import baseConfig from '@polkadot/dev/config/eslint';
 
 export default [
-  ...baseConfig
+  ...baseConfig,
+  {
+    rules: {
+      'import/extensions': 'off'
+    }
+  }
 ];
--- a/packages/extension-base/package.json
+++ b/packages/extension-base/package.json
@@ -39,6 +39,7 @@
     "tslib": "^2.6.2"
   },
   "devDependencies": {
+    "@polkadot/dev-test": "^0.79.3",
     "@polkadot/extension-mocks": "0.48.3-1-x"
   }
 }
--- a/packages/extension-base/src/background/handlers/Extension.spec.ts
+++ b/packages/extension-base/src/background/handlers/Extension.spec.ts
@@ -4,7 +4,7 @@
 /* global chrome */
 
 import '@polkadot/extension-mocks/chrome';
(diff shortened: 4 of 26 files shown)


package.json
```diff
@@ -39,8 +39,10 @@
   },
   "devDependencies": {
     "@polkadot/dev": "^0.79.3",
+    "@types/jest": "^29.5.12",
     "@types/node": "^20.10.5",
     "i18next-scanner": "^4.4.0",
+    "jest": "^29.7.0",
```
TarikGul, line 45: Why are we adding jest?

ryanleecode: compiler failing with `: Cannot find name 'it'. Do you need to install type definitions for a test runner? Try `npm i --save-dev @types/jest` or `npm i --save-dev @types/mocha`.` after i changed the node resolution

TarikGul: I think this is causing problems in others files with the defined custom global jest types referenced in `@polkadot/dev-test`.

TarikGul:
What types do you specifically need exported?

ryanleecode:
> What types do you specifically need exported?

Trying to import @substrate/smoldot-discovery` in `packages/extension`, but i'm getting

```
Cannot find module '@substrate/smoldot-discovery' or its corresponding type declarations.
  There are types at '/home/ryan/Documents/Repositories/polkadot-js/extension/node_modules/@substrate/smoldot-discovery/dist/index.d.mts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.ts(2307)
  ```

ryanleecode:
I can try scoping the moduleResolution update to just `packages/extension`

TarikGul:
> > What types do you specifically need exported?
> 
> Trying to import `@substrate/smoldot-discovery` in `packages/extension`, but i'm getting
> 
> ```
> Cannot find module '@substrate/smoldot-discovery' or its corresponding type declarations.
>   There are types at '/home/ryan/Documents/Repositories/polkadot-js/extension/node_modules/@substrate/smoldot-discovery/dist/index.d.mts', but this result could not be resolved under your current 'moduleResolution' setting. Consider updating to 'node16', 'nodenext', or 'bundler'.ts(2307)
> ```

I think this is an issue with smoldot-discovery and should be solved at that level no? Instead of completely changing the module resolution for one package in the extension? I would be weary of the potential downstream affects it would have since many wallets and users leverage the extension-* packages but additionally the overhead between `@polkadot/dev-ts` vs importing Jest. I just want to ensure we don't add any complexity to the already complex pjs repos.

In terms of adding `@substrate/smoldot-discovery` I am a little out of the loop on that - I think it would be a very large breaking change if it required making webrequests.

ryanleecode:
Should be good now and have no side effects.

TarikGul:
Nice changes, Thank you!

One last thing I would want to be sure of is the potential downstream affects of changing `node` -> `Bundler`. Do we know of any issues it could cause?

ryanleecode:
> Nice changes, Thank you!
> 
> One last thing I would want to be sure of is the potential downstream affects of changing `node` -> `Bundler`. Do we know of any issues it could cause?

The code we are emitting is the same so I think its fine; `moduleResolution` just affects how we resolve imports. `module` has an impact on what is emitted but we are already on `EsNext`.

And since we are using rollup, moduleResolution bundler is the way to go.

https://www.totaltypescript.com/concepts/option-module-must-be-set-to-nodenext-when-option-moduleresolution-is-set-to-nodenext#youre-using-the-wrong-moduleresolution

TarikGul (approved):
Thank you!

PR #1388, opened 2024-07-08
fix: Make extension using Manifest V3 compatible with Firefox

bee344:
This PR's goal is to make the extension compatible with Firefox, while still being compatible with the new `Manifest V3` introduced on #1367 .

#### Changes

- Now there are two manifests: `manifest_firefox.json` and `manifest_chrome.json`, with each containing the configuration for each browser. There's also a dummy `manifest.json` which contents are modified depending on what's the target build. This is set like this to avoid the risk of getting an error in the build after modifying the name of the file and getting stuck. This way if the build is cut short, the only thing that changes is the content of the dummy manifest.
- The target `.zip` files names were updated to reflect the new targets: `master-<TARGET_BROWSER>-build.zip` and `master-<TARGET_BROWSER>-src.zip`.
- The `build` and `build:zip` were updated to run the specific build scripts that were added for each target, in order to not break the CI, but now we can also build just the assets specific for the target browser we want or for both.
- Updated `README.md`


packages/extension-base/README.md
```diff
@@ -5,6 +5,9 @@ Functions, classes and other utilities used in `@polkadot/extension`. These incl
 - message passing,
 - scripts injected inside pages.
 
-They are primarily meant to be used in `@polkadot/extension`, and can be broken without any notice to cater for `@polkadot/extension`'s needs.
+They are primarily meant to be used in `@polkadot/extension`, and can be broken
```
TarikGul, line 8: We can revert these changes :)

TarikGul (approved):
GH_RELEASE_FILES in CI needs to be updated :)

ryanleecode:
running `yarn build` updates the `packages/extension/manifest.json` to `"service_worker": "background.js",`

ryanleecode:
Ah nvm thats intended

TarikGul:
yea `yarn build` is mainly intended for the CI, for browser specific builds run `yarn build:ff` or `yarn build:chrome`

PR #1390, opened 2024-07-11
[DRAFT] feat: extension-rpc

ryanleecode:
Extract the RPC methods outside of `extension-ui` so they can be used elsewhere i.e. the content script.


TarikGul:
cc: @Tbaut

Tbaut:
Can I ask why this PR? What's the benefit of doing this, or why is it needed?

ryanleecode:
i need to access the list of accounts and createTx function in the content script / inpage. 

note: this is still a draft and not final yet; might even scrap this all together

Tbaut:
But can you tell why? This is not linked to any issue, so there's no background info about why this change is needed.

ryanleecode:
We won't be needing this. 

short answer for why is we are testing out the [@substrate/discovery protocol](https://github.com/paritytech/substrate-connect/tree/main/packages/discovery). There will be a forum post explaining this in the near future.

PR #1397, opened 2024-07-14
Add LedgerGeneric support to the extension [WIP]

TarikGul:
WIP

This PR is going to require adding a new `getRawMetadata` interface to the extension in order to work with the new ledger metadataHash proof signing.

My goal is to release the extension first to ensure we can get ledger support in apps. Then work on this immediately after.

### DIFF
--- a/packages/extension-ui/src/hooks/useLedger.ts
+++ b/packages/extension-ui/src/hooks/useLedger.ts
@@ -1,19 +1,16 @@
 // Copyright 2019-2024 @polkadot/extension-ui authors & contributors
 // SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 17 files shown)


packages/extension-ui/src/hooks/useLedger.ts
```diff
@@ -55,7 +52,9 @@ function retrieveLedger (genesis: string): Ledger {
 
   assert(def, 'There is no known Ledger app available for this chain');
 
-  ledger = new Ledger('webusb', def.network);
+  assert(def.slip44, 'Slip44 is not available for this network, please report an issue to update this chains slip44');
```
TarikGul, line 55: The slip44 should always be polkadot for the `LedgerGeneric` wrapper

Issue #1395, opened 2024-07-14
Background task with manifest v3 not working on chrome


poka-IT:
Since this MR: https://github.com/polkadot-js/extension/pull/1367

I'm not able to start the extension after build (with no errors on build), for chrome, testing on Brave.

![image](https://github.com/user-attachments/assets/bc20ca1e-6a6a-4d5b-b3d3-207c5e4d920c)

the logs:

```
index.html:16 Unchecked runtime.lastError: Could not establish connection. Receiving end does not exist.
extension.js:1503 WINDOW;  Window
locales/default/translation.json:1 
        
        
       Failed to load resource: net::ERR_FILE_NOT_FOUND
extension.js:1503 WINDOW;  Window
extension.js:1503 WINDOW;  Window
Error: Attempting to use a disconnected port object
    at extension.js:2:1235104
    at new Promise (<anonymous>)
    at zh (extension.js:2:1235002)
    at Fh (extension.js:2:1235375)
    at extension.js:1503:1979
    at ac (extension.js:2:232469)
    at Cl (extension.js:2:252591)
    at extension.js:2:249195
    at k (extension.js:2:269898)
    at MessagePort.I (extension.js:2:270430)
extension.js:1503 WINDOW;  Window
```

Am I the only one with this issue?

TarikGul:
For chrome itself, we now run `yarn build:chrome` locally. `yarn build` is mainly for the `ci` (I will change this in a follow up PR to be more clear). When you run `yarn build`  it will resolve to the firefox manifest since it's more for testing each build instead of for a specific browser, and firefox is the last step.

<img width="559" alt="Screenshot 2024-07-13 at 9 04 07 PM" src="https://github.com/user-attachments/assets/ddad245f-95f2-422b-8c17-31b9f2aa0032">

poka-IT:
You are absolutly right, that was my mistake...
As we says, RTFM.

The worst is that I had read the new package.json and observed the build first for Chrome then overwritten by Firefox, but it didn't interest me that much...

So I confirm it's working on Brave ^^

poka-IT:
@TarikGul Since `/packages/extension/manifest.json` is overwritten by chrome and firefox manifest on builds step, maybe we should add it to `.gitignore`?

TarikGul:
> Since packages/extension/manifest.json is overwritten by chrome and firefox manifest on builds step, maybe we should add it to .gitignore?

[...]

PR #1398, opened 2024-07-14
Clean manifest build

TarikGul:
This removes the `manifest.json` since `manifest_chrome.json`, and `manifest_firefox.json` are the main sources of config. The build process depending on which browser it is makes a copy of the browser specific config, and sets it as `manifest.json` so there is no need for a template.


bee344 (approved):
Lgtm

TarikGul:
rel: https://github.com/polkadot-js/extension/issues/1395

PR #1408, opened 2024-07-23
feat: Add support for the Ledger Generic App

bee344:
It adds `signsWithMetadata` to support the new Polkadot Generic App. It also adds the support to send the `signedTransaction` back from the extension to the dApp.

Now the option to upload the metadata to the extension also uploads the `rawMetadata` as a `HexString`, that's necessary for `CheckMetadataHash`. The button also remains disabled until the `rawMetadata` is retrieved from the endpoint, in order to make sure the upload is done correctly.

This replaces #1397 .

Needs to be ran alongside [#10778](https://github.com/polkadot-js/apps/pull/10777), because that has the support to uploading the metadata to the extension.

EDIT:
- Added an error when trying to sign with a Ledger without the metadata present in the extension.
- Updated the import process, because it broke with the Generic App and you could no longer import Ledger accounts other than Polkadot's.


packages/extension-ui/src/Popup/Signing/LedgerSign.tsx
```diff
@@ -18,15 +22,35 @@ interface Props {
...
   genesisHash?: string;
-  onSignature?: ({ signature }: { signature: HexString }) => void;
-  payload?: ExtrinsicPayload;
+  onSignature?: ({ signature }: { signature: HexString }, signedTransaction: HexString) => void;
+  payload?: SignerPayloadJSON;
   setError: (value: string | null) => void;
 }
 
+function getMetadataProof (chain: Chain, payload: SignerPayloadJSON) {
+  const m = chain.definition.rawMetadata || '0x00';
```
TarikGul, line 31: I think we should rework this little portion, when no metadata is present there should be some sort of error that is thrown, and logged on the screen.

packages/extension-ui/src/hooks/useLedger.ts
```diff
@@ -102,7 +106,7 @@ export default function useLedger (genesis?: string | null, accountIndex = 0, ad
     setError(null);
     setWarning(null);
 
-    ledger.getAddress(false, accountIndex, addressOffset)
+    ledger.getAddress(chainInfo?.ss58Format || 0, false, accountIndex, addressOffset)
```
TarikGul, line 109: Maybe we can do an `assert` here - We just want to make sure the ss58 is always for certain what is present in the ss58 format and can't resolve to `0`

TarikGul (changes requested):
There are 2 little nits I have - But I also think we should have some way to warn about no metadata being present when users are using their ledger account

TarikGul:
Also we should consider what the workflow is also going to look like for the migration app.

TarikGul (approved):
LGTM, great job - this will be followed up with support for the Migration App.

bee344:
@TarikGul Currently you have to add a new account for every chain, even though the derivation paths are the same, so it's kind of annoying. We could eventually add a "Allow use on any network" option like when creating via mnemonic.

Issue #1411, opened 2024-07-24
Subscribed Accounts don't update when authorizations are added/removed


F-OBrien:
* **I'm submitting a ...**

  - [x] Bug report - (maybe)
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

**Current behavior** - Modifying the authorizations for a connected site does not update the subscribed accounts until either a new key is added/removed from the wallet or the connected page is reloaded.

**Expected behavior** - Modifying connected key authorizations updates the subscribed accounts.

I'll be submitting a PR for consideration.

* **What is the motivation for changing the behavior?**

Subscribed keys not updating immediately can be confusing for users.

* **Please tell us about your environment:**

  - Version: 0.49.3
  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript
    - [ ] Other

Mazaltov888:
Good afternoon. Help me, please. I can't log in to my account. Haven't used the account in a year. Now I downloaded the app in Chrome. Entered the keys, password. I click on the activate button, but it won't click. You can see from the screenshots. Thank you very much in advance for your respect

![WhatsApp Image 2024-07-25 at 15 28 35](https://github.com/user-attachments/assets/447afd32-1a16-4460-9c5d-4a559a521320)

![WhatsApp Image 2024-07-25 at 15 28 50](https://github.com/user-attachments/assets/a3b4cc0d-89b6-4bca-ae9a-b98ed1456d52)

![WhatsApp Image 2024-07-25 at 15 29 15](https://github.com/user-attachments/assets/8b8f20a4-f442-430b-8719-b1ca68525a62)

![WhatsApp Image 2024-07-25 at 15 29 30](https://github.com/user-attachments/assets/7caa1221-2735-4809-beb5-c0092c7d5581)

PR #1412, opened 2024-07-24
Update subscribed accounts when connected site authorizations are modified

F-OBrien:
Addresses #1411

### DIFF
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -529,7 +529,7 @@ export default class Extension {
   }
 
   private updateCurrentTabs ({ urls }: RequestActiveTabsUrlUpdate) {
-    this.#state.udateCurrentTabsUrl(urls);
+    this.#state.updateCurrentTabsUrl(urls);
   }
 
   private getConnectedTabsUrl () {
--- a/packages/extension-base/src/background/handlers/State.ts
+++ b/packages/extension-base/src/background/handlers/State.ts
@@ -284,6 +284,13 @@ export default class State {
     };
   };
 
+  /**
+ * @deprecated This method is deprecated in favor of {@link updateCurrentTabs} and will be removed in a future release.
+ */
+  public udateCurrentTabsUrl (urls: string[]) {
+    this.updateCurrentTabsUrl(urls)
+  }
+
   public updateCurrentTabsUrl (urls: string[]) {
     const connectedTabs = urls.map((url) => {
       let strippedUrl = '';
--- a/packages/extension-base/src/background/handlers/State.ts
+++ b/packages/extension-base/src/background/handlers/State.ts
@@ -154,6 +154,8 @@ export default class State {
 
   public readonly signSubject: BehaviorSubject<SigningRequest[]> = new BehaviorSubject<SigningRequest[]>([]);
 
+  public readonly authUrlSubjects: Record<string, BehaviorSubject<AuthUrlInfo>> = {};
+
   public defaultAuthAccountSelection: string[] = [];
 
   constructor (providers: Providers = {}) {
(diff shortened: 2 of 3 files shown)


F-OBrien:
I you want a UI for testing that uses subscribed accounts you can use https://portal.polymesh.live/ (Polymesh Portal - Testnet)

TarikGul:
Thanks for the PR on the queue to review tomorrow!

F-OBrien:
Side note (this might be a bug?) I did not see any restriction in the signing flow for authorized accounts for a specific URL. Authorized accounts are only checked/filtered out when querying or subscribing to accounts. As a result as long as a key is in the keyring any authorized site can trigger the extension to prompt for signing with it. It just might not have an easy way to know it's in the keyring. I noticed this when I removed an authorization for a key with the current extension release 0.49.1 but was still able to sign with it.

I didn't include it in this PR as I'd be concerned any changes to the signing flow could be breaking for downstream consumers of these packages.

bee344 (approved):
Looks good on my side, thanks. A note is that the renaming will break stuff downstream, so we need to mention it in the next release's notes.

packages/extension-base/src/background/handlers/State.ts
```diff
@@ -271,7 +284,7 @@ export default class State {
     };
   };
 
-  public udateCurrentTabsUrl (urls: string[]) {
+  public updateCurrentTabsUrl (urls: string[]) {
```
TarikGul, line 287: Just as a side note: When we release this - this should be mentioned as a BREAKING CHANGE - it's quite minor but nonetheless will be nice to have in the CHANGELOG and release message. I do like this though.

TarikGul:
> Side note (this might be a bug?) I did not see any restriction in the signing flow for authorized accounts for a specific URL. Authorized accounts are only checked/filtered out when querying or subscribing to accounts. As a result as long as a key is in the keyring any authorized site can trigger the extension to prompt for signing with it. It just might not have an easy way to know it's in the keyring. I noticed this when I removed an authorization for a key with the current extension release 0.49.1 but was still able to sign with it.

@bee344 @ryanleecode @IkerAlus Can you guys help look into this - I think it would important to get some second eyes and clarity on this. :)

packages/extension-base/src/background/handlers/State.ts
ryanleecode: should preserve the original method and just mark it as deprecated and add the new one with the fixed name

packages/extension-base/src/background/handlers/Extension.ts
```diff
@@ -529,7 +529,7 @@ export default class Extension {
   }
 
   private updateCurrentTabs ({ urls }: RequestActiveTabsUrlUpdate) {
-    this.#state.udateCurrentTabsUrl(urls);
+    this.#state.updateCurrentTabsUrl(urls);
```
ryanleecode, line 532: should be the inverse. this function becomes the actual implementation. the misspelled function just calls this one

packages/extension-base/src/background/handlers/State.ts
F-OBrien: I've reintroduced `udateCurrentTabsUrl` and marked as deprecated.

Tbaut:
> As a result as long as a key is in the keyring any authorized site can trigger the extension to prompt for signing with it. 

The idea about the authorization is to make sure 1) a user actually agrees to share their accounts and 2) they can choose which accounts they share (released recently). What you describe certainly sounds like a check that has been forgotten, as I don't see why we should accept signing request for non-shared account. Regarding the gravity, I wonder how it could be exploited though, as in, a website should trigger many signing requests, with many different accounts, to try to guess what account is present in the keyring.

TL;DR: it should be fixed, but it's not a huge issue unless I missed something

TarikGul:
> > As a result as long as a key is in the keyring any authorized site can trigger the extension to prompt for signing with it.
> 
> The idea about the authorization is to make sure 1) a user actually agrees to share their accounts and 2) they can choose which accounts they share (released recently). What you describe certainly sounds like a check that has been forgotten, as I don't see why we should accept signing request for non-shared account. Regarding the gravity, I wonder how it could be exploited though, as in, a website should trigger many signing requests, with many different accounts, to try to guess what account is present in the keyring.
> 
> TL;DR: it should be fixed, but it's not a huge issue unless I missed something

I'll follow up on this by filing an issue today -  and then we can tackle this eventually - and make it a bit more tolerant.

PR #1414, opened 2024-07-25
Rework of the extension landing page

KarimJedda:
As it stands, this page ranks very high for "Polkadot wallet". 

We'll use this as a distribution channel to guide users to ecosystem wallets linked prominently. 

The developer extension is still accessible after a small hurdle. 

https://github.com/user-attachments/assets/8e320f82-3514-4edd-84b2-59d18782ae89


Tbaut:
IMHO, this is a pretty bad (and ugly) way of doing something that's been discussed for a while, which is to de-rank the pjs products such as `apps` and `extension`, and rename them into something that tells they are dev tools. This should be coordinated though.

KarimJedda:
There's no point at all in trying to make the extension landing page beautiful. It's not like the current landing page is an artistic marvel either. 

The rename effort is completely unrelated to this PR, this is about pointing to ecosystem wallets first, c'est tout.

Tbaut (commented):
> As it stands, this page ranks very high for "Polkadot wallet".

Then let's change this. There are also many ppl who land here bc they click on links, what you propose here is a horrendous UX. Let's fix the problem at the source.

KarimJedda:
From where I'm standing it just seems my design skills gave you some grief and are distracting you from the goal this aims to achieve: default redirection to ecosystem wallets. 

If you have constructive suggestions about what is good UX to achieve that, feel free to push your code. We all have opinions, valid or not, but what matters in the end is code, results and time.

The rename/rebrand will be discussed separately, but meanwhile we don't have to wait for the future to fix the issues we know about today. 

Otherwise, this PR will serve as prime example for why in 2024 people still install the pjs extension by mistake when searching for "Polkadot Wallet" and clicking on  https://polkadot.js.org/extension/ , which I remind you is the 2nd hit on Google, and was so for the past couple of years.

Tbaut:
You don't get it. I'm not talking about your UI (the ugly buttons). I'm talking about the wrong solution to the problem, on top of adding a pretty unconventional UX, clicking twice on a button. This behavior is done in applications for very specific cases, when ppl are about to do something dangerous, e.g delete something irreversibly etc. This is not a good solutions to the problems, as I said.

The problem is stated very clearly in your messages:
> people still install the pjs extension by mistake when searching for "Polkadot Wallet" [which] is the 2nd hit on Google

I'm repeating myself here. This is a ranking issue. Let's solve it, instead of providing a bad UX for all users, the ones who don't know, and the ones who know what pjs extension is for.

A lot of work has been done on the wallet page of Polkadot.network, and this is now rank 1. This was not the case before.
The solution I propose is to kill its ranking, there are many ways to do so that Tommi (AliceUndBob) proposed, I'm not an SEO expert but things like removing keyword go a long way. We can also have a banners to direct users to **the ecosystem page** in case tutorials point to the website directly. Another remark, we certainly don't want to update this page or receive issues with "please add my wallet in the list", and gate keep the 3 favorites although there are more.

KarimJedda:
I respect your opinion, but unless you have an actual solution to propose to this specific problem or the other problem you're conflating with this one, don't further derail this PR. I really don't see the utility in yak shaving the UX of a landing page that people shouldn't land on at all. 

> A lot of work has been done on the wallet page of Polkadot.network, and this is now rank 1. 

It's a lot of work yes, but so far your comments have been using "ugly, bad, horrendous, unconventional" without providing a hint of actionable insight, quite the opposite of being constructive or helpful. This isn't the developer experience we should aim for. 

After all, we should leave the SEO ranking part to the experts. Ranking is something that requires coordinated effort and a lot of time, as other websites still link to the extension page and contribute to the ranking. Certainly something needs to be done about the ranking in a next step and the design can be improved iteratively, it's the beauty of open source code. I'm looking forward to your plan and link to your PR for the ranking ;)  

Meanwhile, I have outlined the steps of that plan in our telegram chat a month ago, am executing it, and welcome your contributions. What is sure is that it's out of scope for this PR, so stop trying to make it about that. This PR is to address redirection of people that land on the landing page, not the way how they get there...

itsbirdo:
Gimmie a few days ;) Will propose an alternative

Tbaut:
Please don't take my words as attacks to you or this PR. I've mentioned it many times by now, my feelings about the design have nothing to do with this PR, and I should have left them aside, you're right. UX matters to me though, and I believe there are better ways to achieve the same goal, see below. You chose to go with a PR, rather than an issue + discussion, so we end up discussing in the PR. A closed Tg channel is certainly not the best way to discuss.

I gave you many areas where I think the solutions should go, please take a moment to see them, they are in my past comments:
- change the name
- change the landing to kill its SEO
- add a banner with link to the polkadot wallet section

Very important also: 
- for all the tutorials out there that link to this page, we shouldn't break them
- keep the links to the chrome & firefox store easily accessible, there are still valid reasons to be willing to use this specific extension. we can do this similarly to what the wallet section does with pjs apps on the polkadot website

I agree things should change, and I'm happy to contribute to the discussion and the code, as I did in the past, once we agree on a path forward.

FELLIX000 (approved):

FELLIX000 (approved):

FELLIX000 (approved):

FELLIX000 (approved):

FELLIX000 (approved):

FELLIX000 (approved):

FELLIX000 (approved):

FELLIX000 (approved):

FELLIX000 (approved):

PR #1422, opened 2024-07-29
feat: add support for Ledger Migration App

bee344:
This allows to select on the extension whether to use the Ledger Generic Polkadot App or the Ledger Migration App:
image

It also displays a warning when signing to alert the user of which app it's selected:
image


bee344:
There's this small space at the bottom that shows up when displaying the settings that is really annoying:
image

TarikGul (approved):
Amazing job, lgtm

PR #1424, opened 2024-07-30
fix(bug): extension stuck in `... loading ...` screen after `service_worker` got terminated

bee344:
This sends a message to wake up the `service_worker` when it has been terminated by the spec of the `manifest_V3`. Up to now, after 30 seconds have passed idle, the `service_worker` got terminated, and trying to access the popup would show `... loading ...`, and the user would have to close it and reopen it in order to correctly get the extension to render.

### DIFF
--- a/packages/extension-ui/src/messaging.ts
+++ b/packages/extension-ui/src/messaging.ts
@@ -30,33 +30,34 @@ interface Handler {
 
 type Handlers = Record<string, Handler>;
 
-
-async function wakeupBackground(): Promise<Error | null> {
+async function wakeupBackground (): Promise<Error | null> {
   try {
-    await chrome.runtime.sendMessage({ type: "wakeup" })
-    return null
+    await chrome.runtime.sendMessage({ type: 'wakeup' });
+
+    return null;
   } catch (cause) {
-    return cause instanceof Error ? cause : new Error(String(cause))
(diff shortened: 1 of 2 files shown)


packages/extension-ui/src/messaging.ts
```diff
@@ -30,7 +30,37 @@ interface Handler {
...
+
+    const port = chrome.runtime.connect({ name });
+
+    return port;
+  }
+
+  throw new Error('Failed to create port after multiple attempts', { cause: lastError });
+}
+
+const port = await createPort(PORT_EXTENSION, 5, 1000);
```
TarikGul, line 63: One question: Whats the significant of the delay value being 1000ms and the attempts being 5?

bee344: Well the idea is that this wakes up the service_worker that's been terminated. If the "wake up" messsges are too many, there's something that's not working as it should, and if they are too far from eachother, it doesn't feel smooth, but those values are up to discussion

TarikGul: Yea this makes sense!

TarikGul: Happy to keep the values as is.

TarikGul (approved):
LGTM - great job!

Issue #1432, opened 2024-08-01
Attempting to use a disconnected port


KarimJedda:
## I'm submitting a ...

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

### Current behavior 

Attaching a hopefully descriptive reproduction scenario provided by the Hydration team. 

![photo_2024-07-24_13-44-38](https://github.com/user-attachments/assets/c8800349-74f2-4b37-8a56-51aa1a49080d)

https://github.com/user-attachments/assets/6e4f10bf-9e08-4639-9fac-2eeb65f5297f

### Expected behavior

No `disconnected` issue. There seems to be no error using other ecosystem wallets.

## Misc info 

  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

cc @lolmcshizz

TarikGul:
This could very well be fixed with the newest release of the extension - via https://github.com/polkadot-js/extension/pull/1424

🤞

TarikGul:
This should be fixed with the most recent release on Chrome, do you think you can check and confirm?

KarimJedda:
Checking with the team and will report back, thank you Tarik!

ap211unitech:
Closing this issue since no further clarification has been provided. I hope the problem has been resolved.

Issue #1420, opened 2024-07-28
Polkadot account creation button does not work!


Sunshine-pol:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**

  - Version:
  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

Sunshine-pol:
I use google translate.  Trying to create a polkadot account, when the window opens, I click on create an account, new window to put my name and password twice. then I click the button at the bottom and nothing happens. I use Chrome as my browser.

Sunshine-pol:
dont have yet!

bee344:
@Sunshine-pol 
- Are there any errors or warnings, either in the console or in the popup itself?
- Could you try with the extension in a full window instead of the popup?
- What's your Chrome version?

Sunshine-pol:
Bonjour Hello Alerto Nicolas,

Voici :

   - Are there any errors or warnings, either in the console or in the
   popup itself? no
   - Could you try with the extension in a full window instead of the
   popup? i join one picture
   - What's your Chrome version? the last one. I use of Google Translate:
   And i installed FIREFOX and reproduced the steps: installed the Polkadot
   connection in the toolbar at the top, click on it and start creating a new
   account. To see the photo.

when I click on the orange button at the bottom, nothing happens...as
if the button was not connected!



Merci
Best Regards

Daniel

@Sunshine-pol




*Lavage de vitres - Entretien résidentiel" La qualité attire la qualité"*



*André Daniel St-Arnault*
*Tél: +1-514-952-8688*

***@***.***

<http://goog_1619236767>
www.entretiensunshine.ca



Avis de confidentialité: Ce message électronique, y compris toute pièce
jointe, est privilégié et confidentiel. Toute utilisation ou communication
non autorisée est interdite. Si ce message vous est parvenu par erreur,
veuillez en aviser l'expéditeur par téléphone ou par courriel, supprimer ce
message et toute pièce jointe, et détruire toute copie immédiatement. Merci.

Notice of Confidentiality: This email communication and any attachments are
confidential and legally privileged. Any unauthorized use or disclosure is
[...]

Issue #1417, opened 2024-07-26
Deriving Account Doesn't Work


joepetrowski:
I tried deriving a new account from an existing one. The extension asked for the password of the root account and the derivation path, and when I clicked "Create a derived account", it just hangs

<img width="574" alt="image" src="https://github.com/user-attachments/assets/a85b2f4d-7efe-4bc3-8b5d-96f2f2475f63">

F-OBrien:
It works for me but I'm suspecting this (and other similar issues) are related to ports timing out and disconnecting following the Manifest V3 update as mentioned in #1416. You could try restarting the extension to see if it fixes it.

joepetrowski:
Actually it ended up working when it was just the extension pulled down, but when I made it full screen it doesn't work.

bee344:
@joepetrowski is this on Chrome or FF or both?

joepetrowski:
Have only tried on Chrome.

TarikGul:
This should now be fixed on chrome. Do you mind confirming it works for you?

joepetrowski:
All good, yes. Thanks!

Issue #1416, opened 2024-07-26
can not connect to `polkadot.js` extension


2nthony:
* **I'm submitting a ...**

  - [x] Bug report
  - [ ] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

* **What is the motivation for changing the behavior?**

* **Please tell us about your environment:**

  - Version: 0.49.1
  - Environment:

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [x] JavaScript
    - [ ] TypeScript (include tsc --version)
    - [ ] Other

---

## bug

since the extension upgraded to 0.49.1 this issue occurred

[Subsquare(web)](https://polkadot.subsquare.io) browser console
<img width="407" alt="image" src="https://github.com/user-attachments/assets/cb7ae9a9-a56f-4d69-a81e-2b088d4f47f7">
[polkassembly(web)](https://polkadot.polkassembly.io/opengov) console
<img width="614" alt="image" src="https://github.com/user-attachments/assets/e3919c0b-310d-43eb-92ce-1f68266bbae1">

## repro

- open one of the webs above
- and wait for a while, basically 10 minutes
- wait until the error log appear, and then connect to `polkadot.js`

## code

in subsquare code is using `injectedWeb3['polkadot-js']`

```js
// https://github.com/opensquare-network/subsquare/blob/main/packages/next-common/hooks/connect/useSubstrateAccounts.js#L42-L46

const walletExtension = await injectedWeb3['polkadot-js'].enable("subsquare");
```

F-OBrien:
I've replicated this also. Seems to be the same as the issue described here
https://stackoverflow.com/questions/78165826/chrome-extension-attempting-to-use-a-disconnected-port-object-manifest-v3
and here
https://groups.google.com/a/chromium.org/g/chromium-extensions/c/BD6Tb_AEXfM

2nthony:
seems normal now

PR #1433, opened 2024-08-01
Fix: ensure the service worker is awake before every port message

F-OBrien:
This builds on the approach taken in #1424 but instead of only doing it once when opening the popup it does it before any message over a port from either the extension UI or content script. This prevents attempting to send message over disconnected ports or while the service worker is idle or terminated and **should** hopefully close multiple open issues including #1432, #1420, #1417, #1416, #1415.

This PR does not add a heartbeat to the service worker so if there are transactions that may take a long time to return it is still possible the service worker may timeout and become idle however I've not seen instance of this from my testing as when connected to a dApp there is a ping message being sent every 5 seconds.

secondly I added a check filter out non http url when setting an active tab as the stripUrl function was always erroring for pages like browser tabs like chrome://newtab/ and removed a WINDOW console log which appeared to have been left in when debugging.

@TarikGul

### DIFF
--- a/packages/extension-base/src/utils/portUtils.ts
+++ b/packages/extension-base/src/utils/portUtils.ts
@@ -1,0 +1,62 @@
+// Copyright 2019-2024 @polkadot/extension-base authors & contributors
+// SPDX-License-Identifier: Apache-2.0
+
+import type { Message } from '@polkadot/extension-base/types';
+
+import { chrome } from '@polkadot/extension-inject/chrome';
+
+export function setupPort (portName: string, onMessageHandler: (data: Message['data']) => void, onDisconnectHandler: () => void): chrome.runtime.Port {
+  const port = chrome.runtime.connect({ name: portName });
+
+  port.onMessage.addListener(onMessageHandler);
+
+  port.onDisconnect.addListener(() => {
+    console.log(`Disconnected from ${portName}`);
+    onDisconnectHandler();
+  });
+
+  return port;
+}
+
+export async function wakeUpServiceWorker (): Promise<{ status: string }> {
+  return new Promise((resolve, reject) => {
+    chrome.runtime.sendMessage({ type: 'wakeup' }, (response: { status: string }) => {
+      if (chrome.runtime.lastError) {
+        reject(new Error(chrome.runtime.lastError.message));
+      } else {
+        resolve(response);
+      }
+    });
+  });
+}
+
+export async function ensurePortConnection (
+  portRef: chrome.runtime.Port | undefined,
+  portConfig: {
+    portName: string,
+    onPortMessageHandler: (data: Message['data']) => void,
+    onPortDisconnectHandler: () => void
+  }
+): Promise<chrome.runtime.Port> {
+  const maxAttempts = 5;
+  const delayMs = 1000;
+
+  for (let attempt = 0; attempt < maxAttempts; attempt++) {
+    try {
+      const response = await wakeUpServiceWorker();
+
+      if (response?.status === 'awake') {
+        if (!portRef) {
+          return setupPort(portConfig.portName, portConfig.onPortMessageHandler, portConfig.onPortDisconnectHandler);
+        }
+
+        return portRef;
+      }
+    } catch (error) {
+      console.error(`Attempt ${attempt + 1} failed: ${(error as Error).message}`);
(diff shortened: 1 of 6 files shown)


TarikGul:
Overall, really solid stuff. Thanks so much - I see just the failing tests - and once that is resolved this can go in!

F-OBrien:
> Overall, really solid stuff. Thanks so much - I see just the failing tests - and once that is resolved this can go in!

@TarikGul I tried to fix that test and I'm honestly struggling to figure out how to get it to work. It won't let me mock an implementation of `ensurePortConnection` to return a mockPort with jest.spyOn and I can't see any other easy way to get this to pass. Any help with the test is appreciated. I'm not going to have much access to a laptop again until next Tuesday

TarikGul:
> > Overall, really solid stuff. Thanks so much - I see just the failing tests - and once that is resolved this can go in!
> 
> @TarikGul I tried to fix that test and I'm honestly struggling to figure out how to get it to work. It won't let me mock an implementation of `ensurePortConnection` to return a mockPort with jest.spyOn and I can't see any other easy way to get this to pass. Any help with the test is appreciated. I'm not going to have much access to a laptop again until next Tuesday

Would love to help out on it! I'll have a look :)

TarikGul:
The error comes from https://github.com/polkadot-js/dev/blob/master/packages/dev-test/src/env/expect.ts#L92-L102 which is part of `toHaveBeenCalledWith`. I know this is obvious but I wanted to note it down.

I looked into to this for an hour or so, and will get back to it soon, but some things to note:

1. The extension-mocks package is a bit out of date. It uses sinon-chrome which is note compatible with v3 - we did some patching for now to make it compatible but there could also be an issue here as well in some way.

2. It seems as if there is an underlying issue in the test with how the listeners are being called or captured.

3. The PJS test wrapper is a bit challenging to work with since its very custom, but also has no documentation. I know parts of it, but let me see if I can find some functionality that may help us debug this.

ryanleecode:
@F-OBrien the issue seems to be the test requires the background script to respond with `wakeup` but it seems like sinon isn't running the background script. So the port is never connected

TarikGul:
> @F-OBrien the issue seems to be the test requires the background script to respond with `wakeup` but it seems like sinon isn't running the background script. So the port is never connected

To add to this too `sinon-chrome` doesn't support V3 either which makes it a bit more difficult. But I guess we could monkey patch it ourselves

F-OBrien:
> @F-OBrien the issue seems to be the test requires the background script to respond with `wakeup` but it seems like sinon isn't running the background script. So the port is never connected

Yes, that's why I was trying to mock `ensurePortConnection` or `wakeUpServiceWorker` with jest.spyOn but they just error. If there is a way with Sinon chrome to mock the `awake` response message then that could would work too. I'd have thought the send message API should be the same for manifest V2 and V3 but I've not dug into Sinon chrome to see what it can do.

The test framework seems quite custom so I'm not exactly clear what the options are. I was thinking I could try use to use Sinon rather than jest to fake functions that use the chrome api it instead to see if we can get that to work. Ideally we'd like to be able to test as much of the code as possible without mocks.

Juanma0x:
We have received a report from a user trying to add a Ledger account to the Polkadot extension. Will it be fixed after this PR is deployed?

image

F-OBrien:
@TarikGul this took me way to long to figure out but I've finally got the test to pass. I needed to create a object to wrap `wakeUpServiceWorker` so I could use `jest.spyOn` and also import from the relative path instead of using the `@polkadot/extension-base` path.

F-OBrien:
> We have received a report from a user trying to add a Ledger account to the Polkadot extension. Will it be fixed after this PR is deployed?

@Juanma0x I believe it should

TarikGul:
> @TarikGul this took me way to long to figure out but I've finally got the test to pass. I needed to create a object to wrap `wakeUpServiceWorker` so I could use `jest.spyOn` and also import from the relative path instead of using the `@polkadot/extension-base` path.

Brilliant! You're amazing thanks so much for this - this was pretty critical, hats off to you! Once this goes in I will bump the PJS deps as all the upstream repos have been updated, then we can go ahead and release this and push it to the store.

TarikGul (approved):
lgtm 🚀

PR #1437, opened 2024-08-07
fix: service worker port

ryanleecode:


packages/extension-ui/src/messaging.ts
```diff
@@ -77,6 +77,12 @@ function sendMessage<TMessageType extends MessageTypes> (message: TMessageType,
 
     handlers[id] = { reject, resolve, subscriber };
 
+    // in the event of a service worker crash, the port's name will be
+    // renamed to something bogus. so we need to unset the port variable.
+    if (!(port?.name === PORT_EXTENSION)) {
```
TarikGul:
```suggestion
    if (port?.name !== PORT_EXTENSION) {
```

packages/extension/src/content.ts
```diff
@@ -30,6 +30,12 @@ window.addEventListener('message', ({ data, source }: Message): void => {
     return;
   }
 
+  // in the event of a service worker crash, the port's name will be
+  // renamed to something bogus. so we need to unset the port variable.
+  if (!(port?.name === PORT_CONTENT)) {
```
TarikGul:
```suggestion
  if (port?.name !== PORT_CONTENT) {
```

F-OBrien:
@ryanleecode do you have something I can refer to where it stays a port can be renamed is the service worker crashes or have you experienced this while testing?

If the service worker terminates a `port.onDisconnect` event should be emitted which will set the port to `undefined`.
The only time the current code sets a name for the port stored in `messaging.ts` or `content.ts` is after successfully calling `ensurePortConnection`.

packages/extension-ui/src/messaging.ts
ryanleecode: Ffs still asleep :joy:

TarikGul:
This seems more than reasonable - and I think will help resolve the issue we were discussing about. 

Im going to pull down the branch and do a full test

TarikGul:
> @ryanleecode do you have something I can refer to where it stays a port can be renamed if the service worker crashes or have you experienced this while testing?
> 
> If the service worker terminates a `port.onDisconnect` event should be emitted which will set the port to `undefined`. The only time the current code sets a name for the port stored in `messaging.ts` or `content.ts` is after successfully calling `ensurePortConnection`.

I actually experienced this issue last night while doing full runs before the release. I am having a hard time doing a full step by step repro but I was able to get to happen twice.

Here are some pic's:
Screenshot 2024-08-06 at 10 56 31 PM

Screenshot 2024-08-06 at 10 56 04 PM

ryanleecode:
> @ryanleecode do you have something I can refer to where it stays a port can be renamed if the service worker crashes or have you experienced this while testing?
> 
> If the service worker terminates a `port.onDisconnect` event should be emitted which will set the port to `undefined`. The only time the current code sets a name for the port stored in `messaging.ts` or `content.ts` is after successfully calling `ensurePortConnection`.

we don't know its just a guess rn to solve the above screenshot

F-OBrien:
@TarikGul that is a strange error. The port name is actually correct and not something bogus in the errors you attached so I'm not sure why it is erroring at this check. 

```
assert([PORT_CONTENT, PORT_EXTENSION].includes(port.name), `Unknown connection from ${port.name}`);
```

```
const PORT_PREFIX = `${EXTENSION_PREFIX || 'unknown'}-${process.env['PORT_PREFIX'] || 'unknown'}`;

export const PORT_CONTENT = `${PORT_PREFIX}-content`;
export const PORT_EXTENSION = `${PORT_PREFIX}-extension`;
```
where `EXTENSION_PREFIX = 'polkadot{.js}'`
and `0x8eeee0f75a689b99` is the port prefix `PORT_PREFIX: JSON.stringify(blake2AsHex(JSON.stringify(manifest), 64))`
so the actual ports should be `polkadot{.js}-0x8eeee0f75a689b99-extension` or `polkadot{.js}-0x8eeee0f75a689b99-content`

TarikGul:
> PORT_PREFIX:

Yea this is what I was able to dig up last night as well, which is confused me even more. I would have thought it was a one off thing - but I was able to get this error twice.

ryanleecode:
include `[PORT_CONTENT, PORT_EXTENSION]` as part of the logs to see if they differ

TarikGul:
```
PORT_CONTENT:  polkadot{.js}-0x8ecd92f31039f04d-content
PORT_EXTENSION:  polkadot{.js}-0x8ecd92f31039f04d-extension
```

This is what I am getting on chrome which differs from what the error was showing

F-OBrien:
> ```
> PORT_CONTENT:  polkadot{.js}-0x8ecd92f31039f04d-content
> PORT_EXTENSION:  polkadot{.js}-0x8ecd92f31039f04d-extension
> ```
> 
> This is what I am getting on chrome which differs from what the error was showing
> 
> Let me check firefox and see if there was some kind of mishap

I expect the hash could have changed as the version in the manifest file has changed? You could compare the two manifest files in the master-ff-build and master-chrome-build zips
.

TarikGul:
I figured it out!! WOW what a stupid issue. 

Okay @F-OBrien @ryanleecode thank you both for helping narrow this down. 

SO... I noticed from above the hash that chrome was disconnecting from was the firefox manifest hash. Now how did the firefox hash get mixed up in chrome? I load unpacked in chrome - tested there and then rebuilt the manifest for firefox doing `yarn build:ff`, and since chrome makes a reference to the manifest file that has now changed, it throws the errors along with `...loading...`.

Such an odd error, sorry if any time was wasted for either of you! Going to release the extension now as everything looks good to go :)

PR #1441, opened 2024-08-09
Send ping before subscriptions

F-OBrien:
Send a `ping` message to create a port that will be reused for subsequent messages when the popup is opened. If a port doesn't exist already the `Promise.all` for subscriptions will result in multiple parallel requests to create ports which then triggers the onConnect listener multiple times.


TarikGul (approved):
Seems very reasonable, lgtm Thanks 🚀

PR #1444, opened 2024-08-12
Fix SignArea and ToastProvider timeout

F-OBrien:
in PR #1367 `setTimeout` was replaced with `chrome.alarms` in both the `SignArea` and `ToastProvider` components. Manifest V3 only recommends against the use of `setTimeout` in service workers / background scripts as the scripts don't run continuously. For front end UI components where the timeout is only required while the UI component is active `setTimeout` is still fine to use.

Currently with the `ToastProvider` it is trying to create an alarm of 1.5 seconds. The minimum `chrome.alarms` duration is 30 seconds so this is throwing a console error
> Alarm delay is less than the minimum duration of 30 seconds. In packed extensions, alarm "toastAlarm" will fire after the minimum duration.

Also the code is creating and then removing it straight away and is in a useEffect that only runs once when the provider is initialized. As a result when `show` is called the toast message never clears. This can be seen by clicking the copy button to copy a mnemonic when creating a new account. The timeout/alarm would need to be part of the `show` callback like it was originally. I reverted the changes to `ToastProvider` which resolves these issues.

Similarly there are issues in `SignArea` It is creating and removing the alarm immediately and does not have a listener for the alarm so it does nothing. I've fixed this component to correctly use the alarm but it possibly could also have been reverted to using `setTimeout`. For front end component the only benefit of using an alarm would be if the alarm needed to persist after the component unmounted or if the alarm state needed to be shared across components. (I'd fixed this before I realized `setTimeout` would have been fine to use which is why i didn't just revert this file also but still can if you think it might be a better option)

### DIFF
--- a/packages/extension-ui/src/Popup/Signing/Request/SignArea.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Request/SignArea.tsx
@@ -32,7 +32,7 @@ function SignArea ({ buttonText, className, error, isExternal, isFirst, setError
   const { t } = useTranslation();
 
   useEffect(() => {
-    const lockSigner = async () => {
+    const setSignerLockStatus = async () => {
       setIsLocked(null);
 
       try {
@@ -40,13 +40,15 @@ function SignArea ({ buttonText, className, error, isExternal, isFirst, setError
 
         setIsLocked(isLocked);
 
-        await chrome.alarms.create('SIGNER_TIMEOUT', { delayInMinutes: remainingTime / 60000 });
+        if (remainingTime > 0) {
+          await chrome.alarms.create('SIGNER_TIMEOUT', { delayInMinutes: remainingTime / 60000 });
+        }
 
         if (!isLocked) {
           setSavePass(true);
         }
       } catch (error) {
-        console.error('Error locking signer:', error);
+        console.error('Error setting signer lock status:', error);
       }
     };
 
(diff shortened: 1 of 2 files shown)


packages/extension-ui/src/Popup/Signing/Request/SignArea.tsx
```diff
@@ -32,21 +32,23 @@ function SignArea ({ buttonText, className, error, isExternal, isFirst, setError
   const { t } = useTranslation();
 
   useEffect(() => {
-    const lockSigner = async () => {
+    const setSignerLockStatus = async () => {
```
TarikGul, line 35: I think we can revert this file as well :)

TarikGul (commented):
Great catch on this. Thanks so much for another awesome PR. Yea this was definitely a blunder in the migration. My suggestion is we just revert what we can to keep things simple!

packages/extension-ui/src/Popup/Signing/Request/SignArea.tsx
F-OBrien: @TarikGul I've reverted `SignArea` to use `setTimeout` as it was prior to #1367. I probably should have done this in the first place but I had already done the work of fixing the code before I realized the alarms weren't needed.

TarikGul (approved):

PR #1446, opened 2024-08-14
Fix setting rawMetadata as registry metadata

TarikGul:
This fixes a bug where the rawMetadata would reset the registry properties, and set the addresses to `ss58: 42` when metadata was updated. This fixes the following so that it no longer happens.


bee344 (approved):
LGTM

PR #1453, opened 2024-08-30
Reject instead of ignore

### LINKED ISSUE #1391: Enable Selective Wallet Pairing in web3Enable via Web3AccountsOptions
* **I'm submitting a ...**

  - [ ] Bug report
  - [x] Feature request
  - [ ] Support request
  - [ ] Other

* **What is the current behavior and expected behavior?**

Calling `web3Enable` asks for permissions from all available extensions in a user's browser. This can become undesirable when a user has multiple extensions, as all the wallets will prompt an alert to ask for user permissions.

Additionally, our frontend differentiates between wallets. This means that if a user clicks a button to pair with a specific wallet, all the other Polkadot wallets will still ask for permissions.

Would it be possible to add `Web3AccountsOptions` to `web3Enable` so that a developer can programmatically choose which wallet to pair with?

* **What is the motivation for changing the behavior?**

We have a custom alert that allows a user to choose one wallet instead of another. 
With the current behavior, it is impossible to pair only with the selected wallet.

* **Please tell us about your environment:**

  - Version: "0.46.5",
  - Environment:
  "@polkadot/extension-dapp": "0.46.5",
   "@polkadot/api": "10.9.1",
   "@polkadot/keyring": "^12.6.2",
   "@angular/core": "^14.2.0",
   "@angular/cli": "~14.2.6",

    - [ ] Node.js
    - [x] Browser
    - [ ] Other (limited support for other environments)

  - Language:

    - [ ] JavaScript
    - [x] TypeScript "~4.7.2"
    - [ ] Other

Tbaut:
closes #1391 

This is not prefect, but it's a step in the right direction I believe.
When a user closes the authentication popup, nothing changes, the request is still here, it is still available in the extension on click. The icon also still shows that there's a pending "Auth" request.

Users however can now reject an authentication request. The Dapps will receive a "Reject" and no account will be shared. On the extension side, the connection will be made with 0 accounts and the Dapp that has called `web3Enable` will resolve with 0 accounts.

### DIFF
--- a/packages/extension-base/src/background/handlers/Extension.ts
+++ b/packages/extension-base/src/background/handlers/Extension.ts
@@ -524,9 +524,15 @@ export default class Extension {
     return { list: remAuth };
   }
 
-  private deleteAuthRequest (requestId: string): void {
-    return this.#state.deleteAuthRequest(requestId);
-  }
+  private rejectAuthRequest (id: string): void {
+      const queued = this.#state.getAuthRequest(id);
+  
+      assert(queued, 'Unable to find request');
+  
+      const { reject } = queued;
+  
+      reject(new Error('Rejected'));
+    }
 
   private updateCurrentTabs ({ urls }: RequestActiveTabsUrlUpdate) {
     this.#state.updateCurrentTabsUrl(urls);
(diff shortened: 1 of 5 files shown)


Tbaut:
@nikitayutanov please let me know if this helps

PR #1486, opened 2024-12-04
fix: extension does not get injected on page load

### LINKED ISSUE #1475: Injection happen after page load
Since injection happen [asynchronously](https://github.com/polkadot-js/extension/blob/master/packages/extension/src/page.ts) after the page have loaded, if you try to detect whether PJS is installed too early, it might appeared that the extension is not installed.

tien:
Resolves #1475

### DIFF
--- a/packages/extension/src/page.ts
+++ b/packages/extension/src/page.ts
@@ -31,11 +31,5 @@ window.addEventListener('message', ({ data, source }: Message): void => {
   }
 });
 
-redirectIfPhishing().then((gotRedirected) => {
-  if (!gotRedirected) {
-    inject();
-  }
-}).catch((e) => {
-  console.warn(`Unable to determine if the site is in the phishing list: ${(e as Error).message}`);
-  inject();
-});
+inject();
+redirectIfPhishing().catch((e) => console.warn(`Unable to determine if the site is in the phishing list: ${(e as Error).message}`));


tien:
This is similar to Talisman Wallet does injection: https://github.com/TalismanSociety/talisman/blob/6b49bf846bc5934f70e20e72e8fcaf5ac2c3343a/apps/extension/src/page.ts#L53

packages/extension/src/page.ts
```diff
@@ -31,11 +31,6 @@ window.addEventListener('message', ({ data, source }: Message): void => {
...
-  if (!gotRedirected) {
-    inject();
-  }
-}).catch((e) => {
-  console.warn(`Unable to determine if the site is in the phishing list: ${(e as Error).message}`);
-  inject();
-});
+
+inject();
+redirectIfPhishing();
```
TarikGul, line 36: The promise needs to be handled:

```
inject();
redirectIfPhishing().catch((e) => console.warn(`Unable to determine if the site is in the phishing list: ${(e as Error).message}`));
```

could work?

TarikGul (changes requested):
Needs to be linted: `yarn lint`, and the promise needs to be handled.

TarikGul:
I just need to look into this more, I am not sure of the side affects - like what is the consequences of injecting before a redirect is established? It would nullify the return value of the redirect

tien:
I don't think it matter if a redirect happens. Because the user will no longer be on that page anyway.

Tbaut:
This looks ok to me, as redirecting would still be quick enough for users not to have time interact with the website, and in 99.9% of the case, when there's no redirect, it will save the time it takes to call the phishing api which was delaying the injection.

TarikGul:
Yea did some looking into the actual call stack happening, and I don't think there is any residual issues happening! Going to add this to the release that I cut today

TarikGul (approved):

Issue #1458, opened 2024-09-10
Firefox: Extension shows whitescreen


bkchr:
When trying to sign a transaction, the extension window only shows a white screen. At the beginning you see the window, flashing really fast to just a white screen.

![image](https://github.com/user-attachments/assets/eb3a5ae1-84bb-4452-a1c2-163774b7fb96)

(Also trying to open the window via the extension, shows the same result.)

The following error is printed in the console:

```
Uncaught TypeError: C.name is undefined
    MA moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:389
    Ca moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    _s moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    Cc moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    bu moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    vu moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    yu moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    au moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    ou moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    E moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    O moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
[extension.js:389:2875](moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js)
    MA moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:389
    Ca moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    _s moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    Cc moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    bu moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
    vu moz-extension://2d853b4a-544b-4b75-881b-38d1f4b49329/extension.js:2
[...]

PR #1515, opened 2025-03-05
Ensure text is filtered so it doesnt cause white screens

TarikGul:
The following is a very small fix, but the problem causes white screens when text is undefined. 

rel: https://github.com/polkadot-js/extension/issues/1458


TarikGul:
This PR will be followed up small fix where `<something>.name` is undefined.

PR #1562, opened 2025-06-20
Implement rate limiting for sign requests

ap211unitech:
## 📝 Description

This PR introduces a rate-limiting mechanism for signing requests based on origin. The rate limiter has been designed to use minimal memory. Under the current setup, its storage usage should not exceed approximately 1 KB.


TarikGul (approved):
Sounds reasonable! 🚀

Issue #1578, opened 2025-07-23
Issue when trying to sign Teleport with Ledger


LucasGrasso:
Hello, I'm having a issue when trying to teleport some assets to the People Chain in [polkadot.js](https://polkadot.js.org/apps/#/accounts) using the polkadot extension whilst being connected to my Ledger device. Im using the Generic Ledger App in the extension.
Inspecting the console when trying to sign the transaction im getting:
```
extension.js:1365 Uncaught TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
    at GM (extension.js:1365:76668)
    at i (extension.js:1365:76535)
    at extension.js:1365:76990
    at Array.forEach (<anonymous>)
    at GM (extension.js:1365:76978)
    at extension.js:1365:77298
    at Array.forEach (<anonymous>)
    at ZM (extension.js:1365:77286)
    at Object.j [as getProofForExtrinsicPayload] (extension.js:1365:95834)
```
and no sign prompt appears on my Ledger.

PR #1579, opened 2025-07-28
fix: signing extrinsics on ledger device

rajk93:
**Fixes:** #1578

## Description

This PR addresses the following error encountered when signing an extrinsic using a Ledger device:

```
Uncaught TypeError: undefined is not iterable (cannot read property Symbol(Symbol.iterator))
```

The error occurred during a call to [`getProofForExtrinsicPayload`](https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/Popup/Signing/LedgerSign.tsx#L54), where the `ExtrinsicPayload` created [here](https://github.com/polkadot-js/extension/blob/master/packages/extension-ui/src/Popup/Signing/LedgerSign.tsx#L50) was missing required fields:

* `mode`
* `metadataHash`
* `transactionVersion`

These fields are necessary for generating the signing proof. The fix involves updating the `chain.registry.signedExtensions`, which ensures the missing fields are populated when constructing the `ExtrinsicPayload`.

### DIFF
--- a/packages/extension-ui/src/Popup/Signing/Request/index.tsx
+++ b/packages/extension-ui/src/Popup/Signing/Request/index.tsx
@@ -11,7 +11,7 @@ import React, { useCallback, useContext, useEffect, useState } from 'react';
 import { TypeRegistry } from '@polkadot/types';
 
 import { ActionContext, Address, VerticalSpace, Warning } from '../../../components/index.js';
-import { useTranslation } from '../../../hooks/index.js';
+import { useMetadata, useTranslation } from '../../../hooks/index.js';
 import { approveSignSignature } from '../../../messaging.js';
 import Bytes from '../Bytes.js';
 import Extrinsic from '../Extrinsic.js';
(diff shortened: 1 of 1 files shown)


packages/extension-ui/src/Popup/Signing/Request/index.tsx
```diff
@@ -48,6 +48,15 @@ export default function Request ({ account: { accountIndex, addressOffset, genes
...
   const { t } = useTranslation();
+  const chain = useMetadata(genesisHash);
+
+  useEffect((): void => {
+    // When the chain and request are ready, configure the chain's registry.
+    // This will be picked up by LedgerSign.
+    if (chain && !isRawPayload(request.payload)) {
+      chain.registry.setSignedExtensions(request.payload.signedExtensions, chain.definition.userExtensions);
+    }
+  }, [chain, request]);
```
ap211unitech, line 59: Can we not move this logic [here](https://github.com/polkadot-js/extension/pull/1579/files#diff-92f805934bd6d26657bde5547be1d2726bc02bf0d6a4fa9e67bff245c3a3ad61R70) itself? We can avoid multiple `useEffect`. From high level, I think that should also work.

rajk93: I think it would be better to keep them separate because they have different dependencies. Merging them would cause the request payload to be re-processed unnecessarily whenever the chain information updates, even if the signing request itself hasn't changed.

ap211unitech (approved):
Nice job 🚀

TarikGul (approved):
🚀

Issue #1439, opened 2024-08-08
Ensure Generic Ledger Accounts can use "Allow use on any chain"


TarikGul:
Since the Extension now supports both the generic and migration app there is no way to distinquish whether an account has the old bip32path or the new one which has the slip44 for only polkadot. Because of this we currently don't have a concrete way to allow the `Allow use on any chain` functionality for generic app accounts. 

It would be nice to find a solution on how we can identify an account as a generic app ledger account, then allow the use of the generic chain feature.

Nick-1979:
Why not adding a isGeneric field to AccountJson?

LukeWheeldon:
I am unable to add both a Polkadot and a Kusama Ledger account using the same seed, simultaneously. Every time I need to switch chain, I need to reconnect the Ledger and re-add the account. I assume "Allow use on any chain" could probably fix this issue.

F-OBrien:
Implemented in release 0.63.1

PR #1613, opened 2026-03-26
Allow Ledger generic app accounts to sign on any chain

F-OBrien:
## Summary

This PR enables accounts imported from the generic Ledger app to use the existing Allow use on any chain flow.

It addresses the main concern in #1439: once the extension supports both the generic and migration Ledger apps, it is no longer safe to treat every Ledger account as chain-agnostic by default. This change limits allow-any-chain behavior to the generic Ledger app flow, while preserving explicit network binding for the migration and chain-specific flows.

This PR is stacked on top of [#1611](https://github.com/polkadot-js/extension/pull/1611) and builds on its Ledger transport/signing validation changes.

## What changed

- allow generic Ledger imports to be saved without a genesis hash
- keep network selection required for chain-specific Ledger imports
- allow the generic Ledger flow to initialize and derive without a selected chain
- allow generic Ledger hardware accounts to edit their chain association
- use payload genesis during signing setup so allow-any-chain Ledger accounts still load the correct metadata and signed extensions for the chain being signed

## Genesis hash behavior

For generic Ledger accounts, the genesis hash selected during import, or later from the account page, is primarily account metadata.

Its main purpose is to support dapps that filter visible accounts using account metadata. It is not the sole source of truth for signing in the allow-any-chain flow.

During signing, the extension uses the payload genesis from the transaction request so the correct chain metadata and signed extensions are applied even when the stored account genesis is null or has been changed for filtering purposes.

## Safety and #1439

The main safety concern with allowing a Ledger account to be used across chains is signing with the wrong account or wrong Ledger app mode.

The primary guard remains the sign-time key mismatch validation introduced in [#1611](https://github.com/polkadot-js/extension/pull/1611). If the account derived from the connected Ledger does not match the account expected by the signing request, the UI surfaces a warning instead of silently proceeding.

This is especially important because we do not currently persist which Ledger app mode was originally used to create an account. If a user later changes the Ledger app mode in extension settings, they can recover the correct derivation path by switching back to the matching migration or chain-specific mode. If the active mode derives a different account, the sign-time mismatch check prevents the request from being treated as valid.

So the model here is:

- generic Ledger app accounts can be configured as allow-any-chain
- migration and chain-specific paths remain explicitly network-bound
- stored genesis mainly supports account filtering in dapps
- signing uses payload genesis for chain-specific metadata setup
- signing validates that the connected Ledger-derived account matches the requested account before proceeding

### DIFF
--- a/packages/extension-base/src/background/types.ts
+++ b/packages/extension-base/src/background/types.ts
@@ -201,7 +201,7 @@ export interface RequestAccountCreateHardware {
   accountIndex: number;
   address: string;
   addressOffset: number;
-  genesisHash: HexString;
+  genesisHash?: HexString | null;
   hardwareType: string;
   name: string;
   type: KeypairType;
--- a/packages/extension-ui/src/Popup/Accounts/Account.spec.tsx
+++ b/packages/extension-ui/src/Popup/Accounts/Account.spec.tsx
@@ -12,6 +12,9 @@ import React from 'react';
 import { act } from 'react-dom/test-utils';
 import { MemoryRouter } from 'react-router';
 
+import { settings } from '@polkadot/ui-settings';
+
+import { SettingsContext } from '../../components/index.js';
 import * as messaging from '../../messaging.js';
 import { flushAllPromises } from '../../testHelpers.js';
 import Account from './Account.js';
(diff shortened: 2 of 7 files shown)


TarikGul (approved):
Thanks for the addition, I had to catch back up on the generic ledger app work, but generally looks great. 🚀

PR #1624, opened 2026-07-31
fix: derive raw vs extrinsic signing from the request channel

patriciobcs:
The signing popup decided whether to show the message view or the transaction view by inspecting fields on the payload (`data`, `blockNumber`), while the background picks the signing behaviour from the message channel the request arrived on. Those two are independent and the payload comes from the dapp, so they can disagree.

This makes `RequestSign` a discriminated union on `channel`, set by the background from the message type. Correlating the discriminant with the payload type means consumers narrow through `isExtrinsicRequest` rather than casting, so the display and the signing path can no longer drift apart. `Tabs.extrinsicSign` also rejects a payload carrying a truthy `data`, since that field has no meaning on that channel.

Also branches the signing render on the request rather than on decoded state. Paging between two queued requests of different shapes could otherwise commit one frame using the previous request's decoded payload.

Tests added under `signer channel` in `Extension.spec.ts`.


bee344 (approved):
LGTM

PR #1626, opened 2026-07-31
0.64.1

bee344:
CI doesn't accept X.x.0, it has to be at least X.x.1


bee344:
Did a manual release due to version matching

